Issue #23, opened 2017-05-02
Translations support


JosefNemec:

TheDarocker:
Highly recommend [OneSky](https://www.oneskyapp.com/) if you haven't found a good crowdsourced translation service yet. Completely free for public projects and works really easily once you get everything integrated into the code itself.

JosefNemec:
Thanks for tip, I will definitively check it out once I get to this.

XploitOcelot:
Any possibility to contribute with an european spanish translation?

JosefNemec:
Once this issue is closed then it will be possible to contribute with translations. I haven't consolidated all texts into one place yet.

L1cardo:
i can help with Chinese translation

JosefNemec:
Released in 3.0

PR #123, opened 2017-10-07
Added German Localization.

caltas:
I added German translation for Playnite. I hope it is fine. :)


JosefNemec:
Thanks! I'm afraid there's plenty of more texts to translate, they are just not included in those xaml files. I haven't manage to move all strings there and lot of them are still "hard coded". I recommend following issue #23, once that's done it's safe to translate :)

caltas:
Okay, then just hit me up if it's ready. :)

JosefNemec:
Translation support is now ready, see https://github.com/JosefNemec/Playnite/wiki/How-to:-Translations how to contribute.

PR #161, opened 2017-10-09
 Updated Definitions.yaml, changed platform names for consistency

By6r9:


JosefNemec:
Thanks for a contribution. Just couple of things:

- squash all commits into one
- rename xenia to Xenia

Also I'll have to update UI slightly because check how it looks when BizHawk is imported and you try to add emulated game :D

bizhawk

By6r9:
You can't squash it with merge? There should be an option somewhere on github. If not I try to do it later.

Issue #171, opened 2017-10-10
Improperly named games when importing from folder


r0zbot:
An imported game that has an executable inside a subdirectory of the root game directory will be named after that subdirectory's name. 

Ex: a game located in `Games/GameName/bin/launcher.exe` will be named _bin_ in my library, which is wrong for a lot of games.

That also means that if I import a folder containing a lot of games I will get lots of games with the same name:
![image](https://user-images.githubusercontent.com/28497910/31364852-fa37b13e-ad3d-11e7-91be-30ad48d4dcab.png)

JosefNemec:
I'll probably change it to take a name from executable instead from folder and add option to immediately rename the game from the list (like when importing emulated games).

JosefNemec:
Released in 2.8

PR #206, opened 2017-10-16
Fix: Improperly named games when importing from folder #171

clintonbale:
Improves automatic naming by looking at the version info from the file and getting the Product Name, this also leaves room for more improvement.


JosefNemec:
Thanks for contributions!

PR #291, opened 2017-12-09
 Update Definitions.yaml

By6r9:


JosefNemec:
Looks good, could you just remove `(MAME)` in MAME's profile names? Seems redundant to me.

By6r9:
Done.

Issue #196, opened 2017-10-15
UPlay detect uninstalled games


Kaneth:
Steam, GOG and Battle.net detect uninstalled games - is it possible to do the same for UPlay and Origin?
I guess your hands may be tied by their platforms making it impossible - but just wanted to raise it nonetheless.

Thanks :D

JosefNemec:
Origin already supports uninstalled games. UPlay is currently not technically possible, maybe in future when Ubisoft changes the platform.

Kaneth:
Origin only appears to detect games that are installed or those that have been uninstalled since the installation of Playnite. I have over 15 games on Origin, but the only 2 showing in PlayNite is the one I've installed and the one I've installed and uninstalled. Expected behavior or bug?

JosefNemec:
Are you sure you have full library sync enabled for Origin and are you logged in? If so then please go to "About Playnite..." menu and upload diagnostics package. Thanks

Kaneth:
Yep, definitely logged into Origin with up to date client. Restarted both clients for good measure and Playnite still only sees the 2 games (1 installed, 1 installed and uninstalled since Playnite installation)

[diag.zip](https://github.com/JosefNemec/Playnite/files/1385546/diag.zip)

clibois:
I was wondering if you did try to use a man-in-the middle proxy to see how uplay application get its data ? Guess there is a hidden service that might be accessible once found no ? I have tried on computer t android app but till now uplay launcher seems crash while others app are correctly sniffed. Don't know why though. I might try some other options however(Transparent Proxy).

JosefNemec:
Yes I have tried that, but they push everything though https and they refuse all connections not using their SSL certificate.

JosefNemec:
What might be a better way start is to look at Uplay web and their account pages. I implemented Origin that way, they use the same APIs on web as in Origin client and it was super easy to debug via browser. 

However last time I checked (like a 6 months ago), Uplay doesn't server any web page with your account library that would actually use any kind of API we could use. But it's worth looking into it again.

thetrueavatar:
for information there is an api I have been able to discover while reading the code of the ubisoftclub android application.
Il already discover how to authenticate on that service and get some precious information:
A simple POST on https://public-ubiservices.ubi.com/v2/profiles/sessions will bring you interesting data. 
POST /v2/profiles/sessions HTTP/1.1
Host: public-ubiservices.ubi.com
[...]

PR #390, opened 2018-01-05
Add import all games for Uplay

Astro83:
Thank you for the effort dedicated to this project. I would like to add import all games for Uplay. Thank you very much!


JosefNemec:
@Astro83 I'm not sure you wanted to create new pull request for this, but Uplay is already requested in #196 
@PhrozenByte You are right, I'll add list of other contributors too.

Issue #423, opened 2018-01-12
Emulation Import Wizard: improvements


By6r9:
* selecting more than one entry (or range) with control and shift
* changing platform for that selection 
* sorting by clicking column headers
* extension column 
* maybe select/unselect all button?
* there is one unused (?) column visible after maximizing window

epeternally:
Some form of mass select is definitely needed. MAME sets can be thousands of games long, dealing with them is pretty unwieldy right now.

By6r9:
I don't think it's worth a separate issue, so i'll note that:
* selecting more than one entry (or range) with control and shift
* sorting by clicking column headers
also applies to "Import Installed or from Folder"

kernaldave:
Would be nice to have a drop box to select a platform and apply it to all games being imported.  Right now if I want to change my Game boy Advanced, from Game Boy, I would have to do every one individually.

JosefNemec:
Should be done after/with #653

JosefNemec:
Done in devel9.

PR #425, opened 2018-01-12
New profiles in Definitions.yaml

By6r9:
Added several profiles to make easier game import when emulator supports more systems.

### DIFF
--- a/source/Playnite/Emulators/Definitions.yaml
+++ b/source/Playnite/Emulators/Definitions.yaml
@@ -67,16 +67,51 @@
       Platforms: [Nintendo GameCube, Nintendo Wii]
       ImageExtensions: [elf, dol, gcm, tgc, wbfs, ciso, gcz, iso, wad, dff]
       ExecutableLookup: ^Dolphin\.exe
+    - Name: Nintendo GameCube
+      DefaultArguments: '--exec="{ImagePath}" --batch'
+      Platforms: [Nintendo GameCube]
+      ImageExtensions: [elf, dol, gcm, tgc, wbfs, ciso, gcz, iso, wad, dff]
+      ExecutableLookup: ^Dolphin\.exe
+    - Name: Nintendo Wii
+      DefaultArguments: '--exec="{ImagePath}" --batch'
+      Platforms: [Nintendo Wii]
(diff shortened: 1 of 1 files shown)


JosefNemec:
What's the point of duplicating profiles with the same configuration (GameBoy and Dolphin for example)? I'm maybe not seeing something, but for example Wii and GameCube profiles look exactly the same to me.

By6r9:
Now:
Adding 50 Gamecube games:
Dolphin>Default> 50 items added with first (Gamecube) platform.
Adding 50 Wii games:
Dolphin>Default> 50 items added with first (Gamecube) platform. You need to click every item and change platform.

After adding profiles:
Adding 50 Gamecube games:
Dolphin>Default or Gamecube> 50 items added with first (Gamecube) or Gamecube platform.
Adding 50 Wii games:
Dolphin>Wii> 50 items added with Wii platform. 

Same problem with every other emulator/core which supports more than one platform. You need a step in import wizard with selecting platform, easy way to change platform for multiple items or both those things.

JosefNemec:
I understand and it makes sense to me if the extension support is different based on platform for example. Otherwise I would rather add improvements mentioned in #423 then duplicate entries in emulator definitions.

By6r9:
Removed unnecessary extensions.

JosefNemec:
I still don't like how we would have duplicate records only to separate them by platform. I fully understand that usability of current import dialog is not ideal, but we should fix that (via #423), instead of introducing these workarounds.

By6r9:
Fine with me, this PR is more of a band-aid, and other solution would release me from writing subprofiles for mednafen and bizhawk.

JosefNemec:
I'm willing to merge this if you make some changes:

* Remove duplicated entries. For example you created new Wii and GameCube profiles, but kept the one that contains both platforms. Just remove the original profile that contains all platforms and keep separate ones.
* Make sure that separate profiles only have file extensions supported by that platform.

By6r9:
- removed
- duplicate extension are either archives, or generic extensions (.bin, for example). For consoles there are usually only one or two typical extension per platforms. Ping me here, or on discord if I should double-check something.

PR #477, opened 2018-02-16
Force HTTPS for BattleNet login window and Steam private account URLs

Shanezlar:
This is necessary for security reasons. 

BattleNet login was easily phishable in an MITM-scenario.

Steam private account requests over HTTP would broadcast apiKey and userName in plain text.


JosefNemec:
Just to be sure, have you tested that all servers actually serve https?

Shanezlar:
Yes, I have tested all the hardcoded domains and currently used dynamic ones (seems to me that in the case of GogLibrary.cs the domain, e.g. for image location, is theoretically dynamic based on the server reply. But the domain currently used by GOG supports https).

Also in my test build all affected functions were working as expected.

Issue #540, opened 2018-04-27
Importing games from D-Fend Reloaded (DOSBOX frontend) doesn't work


spiral6:
The issue is due to typical DOSBOX .conf files being renamed .prof files in D-Fend. If there's an option to import them as .prof files, it would work.

JosefNemec:
You need to create new [emulator profile](https://github.com/JosefNemec/Playnite/wiki/Emulator-Settings) for it then.

And possible if you can also add it into [this list](https://github.com/JosefNemec/Playnite/blob/master/source/Playnite/Emulators/Definitions.yaml) and send pull request for it. That will make supported by default in Playnite.

spiral6:
Actually, the commit I added isn't going to work. When launching a game from D-Fend Reloaded from the command line, it needs the "proper name", aka the name of the game in Playnite (i.e. "Jazz Jackrabbit" as opposed to "jazzjackrabbit.prof" or "jazzjackrabbit") to launch. Perhaps add a {Name} option?

JosefNemec:
I'll add `{Name}` option in next update.

JosefNemec:
There's now new variable called `{Name}` you can use in 4.2.

PR #542, opened 2018-04-29
Added definition for D-Fend Reloaded

spiral6:
for solving issue #540.

### DIFF
--- a/source/Playnite/Emulators/Definitions.yaml
+++ b/source/Playnite/Emulators/Definitions.yaml
@@ -11,7 +11,7 @@
   Website: 'http://dfendreloaded.sourceforge.net/'
   Profiles:
     - Name: Default
-      DefaultArguments: '"{ImageNameNoExt}"'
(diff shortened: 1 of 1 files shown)


source/Playnite/Emulators/Definitions.yaml
```diff
@@ -7,6 +7,15 @@
       ImageExtensions: [conf]  
       ExecutableLookup: ^dosbox\.exe
 
+﻿- Name: D-Fend Reloaded (DOSBOX frontend)
+  Website: 'http://dfendreloaded.sourceforge.net/'
+  Profiles:
+    - Name: Default
+      DefaultArguments: '"{Game}"'
```
JosefNemec: Are you sure this will work? I would use `{Name}` variable introduced in 4.2

spiral6: Changed. Feel free to merge.

PR #617, opened 2018-05-20
FirstTimeStartup UI Polish

spektor56:
This is a WIP

### DIFF
--- a/source/PlayniteUI/Windows/FirstTimeStartupWindow.xaml
+++ b/source/PlayniteUI/Windows/FirstTimeStartupWindow.xaml
@@ -37,7 +37,7 @@
                     <Setter Property="Template">
                         <Setter.Value>
                             <ControlTemplate TargetType="TabControl">
-                                <ScrollViewer IsTabStop="True" x:Name="TabControlScroller" HorizontalScrollBarVisibility="Disabled"  VerticalScrollBarVisibility="Auto" >
+                                <ScrollViewer x:Name="TabControlScroller" HorizontalScrollBarVisibility="Disabled"  VerticalScrollBarVisibility="Auto" >
                                     <Grid>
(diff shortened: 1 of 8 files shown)


JosefNemec:
What about showing user dialog message asking them if they want to skip wizard dialog instead of shutting down?

spektor56:
Is there a reason people will need to skip the first-time setup wizard?  If you don't want to import anything you can just uncheck everything and click next.  If there is then the caption icons should probably not be hidden

JosefNemec:
Probably showing close icon and displaying warning about skipping the wizard might be a good solution.

source/PlayniteUI/Windows/FirstTimeStartupWindow.xaml
```diff
@@ -12,7 +12,7 @@
         Style="{DynamicResource StandardWindowStyle}"
         Title="{DynamicResource LOCFirstWindowTitle}" WindowStartupLocation="CenterScreen"
         ShowCloseButton="False" ShowMaximizeButton="False" ShowMinimizeButton="False"
-        Height="450" Width="550" x:Name="WindowMain"
```
JosefNemec: This makes dialog super huge, please change this to some default value.

spektor56: the height and width definition is in the theme now because it is dependent on the theme.  If the setting is here the theme setting gets overridden.  If there a way like in CSS to force the theme to override I can do that.

spektor56: Yes, moved the sizing back to the form (using the largest of the 2 so both styles work nice by default).  Wasn't thinking it being the "default" fallback for other windows, though they all have explicit sizes set right now and it wasn't applying to them.

PR #693, opened 2018-06-28
Resize Jitter Fix

### LINKED ISSUE #614: Excessive jitter when resizing windows in windows 10
WindowChrome doesn't seem to play nice in windows 10 when NonClientFrameEdges is "None" and GlassFrameThickness is "0".  If NonClientFrameEdges is set to "Bottom" and GlassFrameThickness is anything but "0" it works fine.  If GlassFrameThickness is set to "0" the classic tool menu flashes into focus when resizing.

spektor56:
Fixes #614


JosefNemec:
Couple issues I noticed:
* There's noticeable border around window's content when window is maximized
* Window buttons (maximize, minimize and close) don't work with Classic theme

spektor56:
There has always been a border in the theme style when maximized, maybe a bug was preventing it from displaying properly before?

JosefNemec:
I'm sorry it took me so long to get to this. You are right that the border was there all the time, but it simple doesn't render currently. Check this comparison:

With your fix:
image

Current master:
image

I just want this to look the same as current master (with those lines extending to the border of the screen). Also there's now merge conflict with master.

spektor56:
Always forget to change classic... Already removed the padding from the modern theme.

JosefNemec:
While testing new version I've encounter one more issue with this fix. When new window is opened (main one or any child window) the whole window content is white for a while until content is loaded. It's more visible on slower systems or if you are connected via RDP. On faster systems it can be easily reproduced when you switch from fullscreen mode to desktop one. Can you check if this can be fixed? I would have to revert this otherwise, since it's really distracting.

spektor56:
Think they already have a fix for that, i will merge their changes and rebuild the DLL

PR #706, opened 2018-07-13
Fix scrollviewer style exception in debug console

spektor56:


spektor56:
still need to do the classic one

PR #741, opened 2018-08-31
update ControlzEx

spektor56:


spektor56:
Dont merge yet

JosefNemec:
I think we can cancel this, I already merged updated ControlzEx to devel.

spektor56:
I only see a DLL update you did to the devel branch, what about the other changes?  Where did you get the DLL from?  I have a special branch of controlzEx with the captionheight added back.

Issue #782, opened 2018-10-04
Space gets deleted when entering into filter TextBox


jharmer95:
When typing into a filter textbox, a space is automatically backspaced. Inserting a space between characters instead of at the end still works.
![playnitespaceissue](https://user-images.githubusercontent.com/21198728/46447427-c2d6c900-c74f-11e8-9639-bfcd758a6590.gif)

jharmer95:
Found responsible line.
Line 33 in source/PlayNiteUI/Converters/ListToStringConverter.cs:

```C#
// source/PlayNiteUI/Converters/ListToStringConverter.cs
var converted = stringVal.Split(new char[] { ',' }).SkipWhile(a => string.IsNullOrEmpty(a.Trim())).Select(a => a.Trim());
```

The `Trim()` in the Select query is causing the tailing space to be removed.

Potential change to fix this:

```diff
- var converted = stringVal.Split(new char[] { ',' }).SkipWhile(a => string.IsNullOrEmpty(a.Trim())).Select(a => a.Trim());
+ var converted = stringVal.Split(new char[] { ',' }).SkipWhile(a => string.IsNullOrEmpty(a.Trim()));
```

Cheers.

jharmer95:
Pull request ( #786 ) made.

PR #786, opened 2018-10-09
Fixed accidental trimming of spaces put into filters

jharmer95:
Fixed issue ( #782 ) where trailing space was deleted automatically preventing the entering of filters that contain whitespace (e.g. a Category named "To Be Completed")

### DIFF
--- a/source/PlayniteUI/Converters/ListToStringConverter.cs
+++ b/source/PlayniteUI/Converters/ListToStringConverter.cs
@@ -30,7 +30,19 @@ namespace PlayniteUI
             }
             else
             {
-                var converted = stringVal.Split(new char[] { ',' }).SkipWhile(a => string.IsNullOrEmpty(a.Trim()));
+                IEnumerable<string> converted;
+
+                if (parameter is string s && s.Equals("true"))
+                {
+                    // don't use final trim()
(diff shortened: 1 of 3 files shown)


JosefNemec:
Have you tested that this doesn't break other places where this converter [is used](https://github.com/JosefNemec/Playnite/search?q=ListToStringConverter&unscoped_q=ListToStringConverter)? Or that the search still works properly when trailing white spaces are present? I'm particular skeptical about game edit window still working properly if you don't remove trailing spaces.

JosefNemec:
I don't have a problem with adding parameter controlling whether trimming should be used.

jharmer95:
Added control logic for ListToStringConverter to determine whether a second trim occurs.
Added a ConverterParameter (string "NoTrim") to the Modern.xaml and Classic.xaml files to allow the filter fields to ignore final whitespace trimming, but the game edit and any other windows that use the ListToStringConverter won't be affected unless they are explicitly given a ConverterParameter of "NoTrim"

JosefNemec:
Sorry it took me song long to review this. As far as custom skins goes, that won't be an issue in future since theming API will be reworked and this will be handled differently.

Issue #769, opened 2018-09-23
Steam integration - import of installed source mod games


JosefNemec:
Example how to start a mod:
`steam.exe -applaunch 215 -game "E:\steam\steamapps\sourcemods\half-life 2 riot act"`

JosefNemec:
Released in 4.55

PR #788, opened 2018-10-13
Add support for Source mods in Steam (issue #769)

psychonic:
As asked for in issue #769, this adds support for scanning/launching Source mods on Steam.

To do this, I had to change the Playnite Game ID to use a Steam Game ID instead of App ID, but there shouldn't be any backcompat issues since they have the same value for games that were already supported. (GameID is a 64-bit id, with the AppID as the lower 32 bits, and the high bits unset unless it's a mod or shortcut). There shouldn't be any effective change at all for non-mod games.

Metadata comes from the mod's gameinfo.txt, same as how Steam does it. For mods that have it populated, this provides the mod name, developer name, links (developer + manual), and categories (single-player, multi-player, or both). Icon can also be populated if Playnite can support tgas. (I couldn't find the icon image code, but I just assumed that it does not). Launching mods works as expected.

Let me know if any changes are desired. One thing I wasn't sure how to best handle was detecting installation state and uninstalling. I have a WIP change to detect installation state (making sure both the mod and the base game are installed), but not sure what to do for uninstall. The Steam UI does not allow uninstallation of mods, and I don't know if we want to just be directly deleting files from here. Also, the most that Install could do if the mod was fully uninstalled would be to just install the base game (via Steam), which would not install the mod from scratch.

I tested with a variety of Source mods I had already installed from ages ago, including The Hidden, Research and Developer, GoldenEye: Source, BattleGrounds 2, and SourceForts.

### DIFF
--- a/source/Plugins/SteamLibrary/SteamLibrary.cs
+++ b/source/Plugins/SteamLibrary/SteamLibrary.cs
@@ -127,6 +127,31 @@ namespace SteamLibrary
             return games;
         }
 
+        internal List<Game> GetInstalledGoldSrcModsFromFolder(string path)
+        {
+            var games = new List<Game>();
+
+            var firstPartyMods = new string[] { "bshift", "cstrike", "czero", "czeror", "dmc", "dod", "gearbox", "ricochet", "tfc", "valve"};
+
+            var dirInfo = new DirectoryInfo(path);
+
+            foreach (var folder in dirInfo.GetDirectories().Where(a => Array.IndexOf(firstPartyMods, a.Name) == -1).Select(a => a.FullName))
+            {
+                try
+                {
+                    var game = GetInstalledModFromFolder(folder, ModInfo.ModType.HL);
+                    games.Add(game);
+                }
+                catch (Exception exc)
+                {
+                    // gameinfo.txt may not exist or may be invalid
+                    logger.Error(exc, $"Failed to get information about installed GoldSrc mod from: {path}");
+                }
+            }
+
+            return games;
+        }
+
         internal List<Game> GetInstalledSourceModsFromFolder(string path)
         {
             var games = new List<Game>();
@@ -135,7 +160,7 @@ namespace SteamLibrary
             {
                 try
                 {
-                    var game = GetInstalledSourceModFromFolder(folder);
+                    var game = GetInstalledModFromFolder(folder, ModInfo.ModType.HL2);
                     games.Add(game);
                 }
                 catch (Exception exc)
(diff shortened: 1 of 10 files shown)


JosefNemec:
Haven't had a time to check this out fully yet, but it looks good. Thanks a lot for contributing!

Do you think it would be hard to extend this with support for Half-Life 1 mods as well (I've been playing those a lot lately :)? I know that the detection/installation is quite different compared to source mods, so I'm just wondering if you could do it as part of this or if I should open new issue for it.

Also do you think it would be worth adding option to settings to disable this? Add something like `Import Source and GoldSrc mods` checkbox. Have it enabled by default and just have it as an option for people to disable import of mods. But I'm not sure if it's something people would want to disable.

psychonic:
I took a brief look now at HL1 mods.

Getting the base path for them via the registry is similar, and then iterating child folders to look for liblist.gam files. We'd have to use a separate parser for those, or just regex pattern matching to pull values out. All of the same Game ID should work.

What I'm not clear on it how Steam differentiates between a first-party mod and a third-party mod. For example, tfc has its own liblist.gam there. Using the same logic, we would detect that and then show it installed twice (once as mod, and once as app). I don't have a good solution for that, other than just hardcoding an hl1 mod blacklist for the few first-party mods.

I'll start on a change to get them in.

I don't foresee people wanting to disable mods, at least not enough making it worth the effort to add the option. I would think that if they have it showing up in Steam, they would want it showing up in Playnite.

Do you happen to have any thoughts on the potential issues with install/uninstall state and actions that I had mentioned?

JosefNemec:
I guess there's just a small number of first party mods so some kind of black list here wouldn't hurt.

For the settings option, yeah you're right. If it shows in Steam it should show in Playnite, forget it then.

As for the installation support. I would just display error that install/uninstall is not supported for mods since there's no way how to implement that reliably. We could technically just delete the folder for uninstall, but what if some installer was used to install mod? We would be leaving stuff like registry entries depending on what installation method was used.

psychonic:
I added HL1 mod support now. The changes got a little messier, but it was worse before cleaning it up a bit.

I also added the error dialog for attempting to install or uninstall a mod, but I wasn't sure how to best convey further up that the action has been canceled. (So the "Installing..." or "Uninstalling..." state is still present in the UI until cancelling there).

source/Plugins/SteamLibrary/SteamLibrary.cs
```diff
@@ -125,6 +127,73 @@ internal List<Game> GetInstalledGamesFromFolder(string path)
...
+                {
+                    // gameinfo.txt may not exist or may be invalid
+                    logger.Error(exc, $"Failed to get information about installed Source mod from: {path}");
+                }
+            }
+
+            return games;
+        }
+
+        static internal Game GetInstalledModFromFolder(string path, ModInfo.ModType modType)
```
JosefNemec: Remove static and inject complete` SteamLibrary` into `SteamMetadataProvider` (you can replace `Settings` object since it comes from SteamLibrary`), that way you can also remove `Id` workaround on line 488.

JosefNemec:
Review done, I've also added support for TGA images in `devel_4.5` branch so you can also assign mod icons. Just assign full path to icon file in `GetInstalledModFromFolder` to `Icon` property and Playnite will do the rest (for first time import only).

psychonic:
Thanks for the feedback! I pushed changes for everything but the last note, regarding `SteamLibrary` and `SteamMetadataProvider`. I wasn't completely clear on what you were suggesting. I also haven't fully wrapped my head around how the classes tie together as-is, which is probably part of the issue there.

JosefNemec:
Basically this is how new constructor for metadata provider will look like:
```
public SteamMetadataProvider(SteamServicesClient playniteServices, SteamLibrary library, SteamApiClient apiClient)
{
    this.library= library;
    this.playniteServices = playniteServices;
    this.apiClient = apiClient;
}
```
You will need to replace all usages of `settings` to `library.Settings`.

New injection from `SteamLibrary` will look like this:
```
return new SteamMetadataProvider(servicesClient, this, apiClient);
```

JosefNemec:
One more thing, can we change use of `Categories` to `Tags`? Right now Categories are only user definable and none of the automatic imports set them up, tags are used instead. Just for consistency and to not affect user's categories since you can group the view by them.

psychonic:
I understand now. That makes much more sense. Also, the Categories versus Tags was an overview. I meant to do it the same way that Steam store categories are handled, but the naming threw me off.

I didn't test the previous changes and really should have. I found a couple of bugs with them. Everything is fixed up and tested again now, in addition to all of the suggested changes. Let me know if there's anything else to change. Thanks!

PR #870, opened 2018-12-12
New Russian Translate

ziv808:
Hello! I made a new version of the Russian localization for the application version 4.63.
The translation was checked and translated correctly, the program also behaves stably, without errors. (The first time we translated, after we completed the translation, the program did not want to run).

I would also like to emphasize that in the file “english.xaml” there is no phrase “Export Library” from the section “Extensions”. In addition, I would like to add that I really liked the program, so I will localize this program further!
[russian new.zip](https://github.com/JosefNemec/Playnite/files/2671924/russian.new.zip)


JosefNemec:
I merged the file manually, this PR is obviously wrong since it contains a lot of irrelevant commits.

Issue #837, opened 2018-12-01
Exlcusion filtering


Poggles:
Sometimes you want to see every single game to check for duplicates etc, it would be useful to start typing a title and see games that you have installed/uninstalled AND hidden titles at the same time.

Mates1500:
Is this being worked on? I could take a look at it.

JosefNemec:
I'm not working on it, feel free to look at it.

Poggles:
Depending on whether further (large) changes are already planned for the way filtering works, a quick fix could be to add another checkbox for "everything including hidden".

I really appreciate the work you guys are doing on this.

JosefNemec:
I think the easier solution would be to change the checkbox to three-state one:
- checked: show only hidden
- intermediate: don't show hidden
- unchecked: show hidden as well

That could be done for other filters as well (like Installed, Uninstalled).

Mates1500:
I've created a pull request suggesting consolidation of the entire checkbox logic with some questions still left unanswered, your feedback would be appreciated, please check out #881

Poggles:
I like this suggestion.  It gives me all the functionality I felt was missing in the filtering.  You could even extend this to the textbox filters (a checkbox next to the textbox to indicate if the text input is to be exclusively or inclusively applied).

I was picturing in my head a way of displayed "AND / OR" filtering to users similarly to the filters in isthereanydeal.com but I couldn't think of a way to show it concisely in the current UI.

The use of the different checkbox options seems ideal to me.  I also agree with the "reset filters" and tooltip idea.  I am not certain if the "Enable Inclusive" icon is clear enough without tooltips.  What about some subtle colouring:

1. Disable  - slightly greyed (as in your pic in #881)
2. Enable Inclusive - Normal white
3. Enable Exclusive - Hint of red?

Colour might be excessive to be honest.

jpflanegan:
I have given this much thought and I think I have a workable solution. Not sure how easy it is to implement though. I found showing or not showing hidden games was causing problems with filtering and knowing what would appear or not.

These are the changes I would make

Add a "Show Hidden Games" toggle switch which can be on or off
Change all other checkbox UI functionality so that first click on a checkbox will be a solid checkbox, second click will be a tick, third click will clear it.
If a checkbox has a solid square then a game only has to meet one of the checkbox filters with a solid square to be included in the results
If a checkbox is ticked then the game has to meet all of the checkbox filters with a tick to be included in the results
An example of filter below
[...]

PR #881, opened 2018-12-17
Checkbox filter rework (#837) 

Mates1500:
Ok, so I've thought about the filters for a while (#837), and the first thing that stood out to me was how each checkable filter did have very different behavior from the others. This left me (and very likely many more users) somewhat confused, so I've consolidated the behavior to be CLEARLY readable from its visual state.

Therefore I've reworked the checkbox filter to have 3 possible states

1. Disable (Exclude completely)
2. Enable Inclusive (Include, the condition does NOT have to be fulfilled)
3. Enable Exclusive (Include, the condition HAS to be fulfilled)

Example of all respectively numbered can be seen here 
image

Using common sense, currently the following checkboxes have these states available from the UI

- Installed - Disable, Enable Inclusive
- Uninstalled - Disable, Enable Inclusive
- Hidden - Disable, Enable Inclusive, Enable Exclusive
- Favorite - Disable, Enable Inclusive, Enable Exclusive

Which makes the "Clear Filters" button even a little more confusing.

In the end there are still some unanswered questions left

- Should the Favorite filter even have the "Disable" option available?
- Wouldn't it be more sensible to relabel the "Clear Filters" button to "Reset Filters" which would save the user some clicks when configuring their own filter? If so, what should be the default values? IMO Installed - Enable Inclusive, Uninstalled - Enable Inclusive, Hidden - Disable, Favorite - Enable Inclusive would be a good start.
- We could perhaps include a help tooltip explaining what each checkbox state means instead of forcing the user to do some trial/error?


JosefNemec:
Thanks for the contribution, I'll try to review it no later then by the end of the week.

JosefNemec:
First I would change the behavior of the checkbox state to correspond with other places where we use three state checkboxes (like when settings categories for multiple games):
 - Unchecked - exclude
 - Indeterminate - does not have to match condition
 - Checked - has to match condition

As for other questions:
- I don't see why favorite filter shouldn't offer disable option as well.
- Renaming to "Reset Filters" probably makes more sense. The default state should display all games except hidden ones.
- Definitely include tooltip to explain what each setting does.

Also I noticed that this change will cause issues for people after upgrading. The original type of filter values was `bool` and it's now `ThreeStateFilterEnum`. I'm pretty sure that deserialization of existing settings will fail after upgrade. Another issue is that previous default state makes settings for installed and uninstalled games deselected resulting in no games being shown.

Mates1500:
ok, I think I've addressed all of these things in the commits referenced above

as for the saved settings, I've tried manually altering the filter settings in the json to something like `true, true, false, true` and they simply reset to their default enum values without any additional changes

Mates1500:
Maybe make installed/uninstalled just a two-state checkbox (unchecked; indeterminate) and leave favorites and hidden as-is? 
I agree that the current state it is in, might not be obvious from the get-go, but at least the visual language is clear. If a checkmark on one item does not do the exact same thing as a checkmark on the item below it (which is the pre-pull-request state), then I'd find the behavior very inconsistent.

JosefNemec:
My biggest issue is when you start combining more filters together. Even when speaking only about Hidden and Favorite together, it's not very clear which setting overrides other (at least to me). Also we shouldn't probably treat Favorite and Hidden the same way since they server different purpose (the game fields, not the filter settings).

jpflanegan:
I have given this much thought and I think I have a workable solution. Not sure how easy it is to implement though. I found showing or not showing hidden games was causing problems with filtering and knowing what would appear or not.

These are the changes I would make
1. Add a "Show Hidden Games" toggle switch which can be on or off
2. Change all other checkbox UI functionality so that first click on a checkbox will be a solid checkbox, second click will be a tick, third click will clear it.
3. If a checkbox has a solid square then a game only has to meet one of the checkbox filters with a solid square to be included in the results
4. If a checkbox is ticked then the game has to meet all of the checkbox filters with a tick to be included in the results

An example of filter below
By default you will see your library except for hidden games.
Click Show Hidden Games you will see everything.
Single click Installed making it a solid square you will still see everything.
Single click Uninstalled making it a solid square you will still see everything.
Single click Installed making it a tick will show installed games as well as installed hidden games.
Click Show Hidden Games will show installed games but not installed hidden games.
Single click Uninstalled making it a tick will show nothing.
Click Show Hidden Games will show noting.
Click Installed which will clear it will show uninstalled games as well as uninstalled hidden games.
Click Hidden twice making it a tick will show only hidden and uninstalled games.

I have sat and thought through how selecting different options might return results and this seems to be the best way to have an AND and an OR option together for filtering. If I had the skill to code this to test I would do it but I am still learning. This could be incorporated for each filtering field and should give satisfactory results.

I also feel it fits with the rest of the UI experience when selecting multiple games and want to set a tag or category and only if all of the selected items have that tag will it show a tick otherwise it shows a solid checkbox.

JosefNemec:
I'm closing this since it's now obsolete and ti will have to be completely rewritten for #2538

PR #884, opened 2018-12-19
Update italian.xaml and little problem

tesivo:
247    <sys:String x:Key="LOCSettingsSkinFullscreen">Fullscreen Theme</sys:String>
248    <sys:String x:Key="LOCSettingsSkinColorFullscreen">Fullscreen Theme Profile</sys:String>

Looks like the second line overrides the first line. In the English version the problem does not emerge, while, for example in Italian, there is an overlapping of text since the strings would be:

247  <sys:String x:Key="LOCSettingsSkinFullscreen">Tema Schermo Intero</sys:String>
248  <sys:String x:Key="LOCSettingsSkinColorFullscreen">Colore del tema Schermo Intero</sys:String>


StarFang208:
I wonder how could we fix that, unless we decide to pull the cheap trick of using shortened version of one of the words.

tesivo:
In the program I found only the short string used, where would the other be?

JosefNemec:
I can make the text box to properly resize to fit the text. Please open new issue for it.

tesivo:
Done ;)

PR #904, opened 2018-12-29
Better wizard and manual game addition

romgerman:
#909


source/PlayniteSDK/Plugins/IGameLibrary.cs
```diff
@@ -29,6 +29,12 @@ public interface ILibraryPlugin : IPlugin
         /// </summary>
         string Name { get; }
 
+        /// <summary>
+        /// Gets if this game library provider is installed on PC
+        /// </summary>
+        /// <returns></returns>
+        bool IsInstalled { get; }
```
JosefNemec: Please rename to something like `IsClientInstalled`, I think it would better represent what the property should be used for.

source/PlayniteUI/ViewModels/InstalledGamesViewModel.cs
```diff
@@ -335,7 +336,7 @@ public void SelectExecutable()
             var program = new ImportableProgram()
             {
                 Icon = path,
-                Name = new DirectoryInfo(Path.GetDirectoryName(path)).Name,
+                Name = FileVersionInfo.GetVersionInfo(path).ProductName ?? new DirectoryInfo(Path.GetDirectoryName(path)).Name,
```
JosefNemec: Also check if `ProductName` is not empty string, not just null.

romgerman:
Alright. I've update the pull.

PR #920, opened 2019-01-06
Improve UI interactions in game edit and category windows

bburky:
Game edit window changes:

* Set default button in `GameEditWindow` to `ButtonOK`

Category config window changes:

*  Conditionally set default button in `CategoryConfigWindow`. If `NewTestCat` is not focused, use `ButtonOK` as default. If   `NewTestCat` is not focused, use `ButtonAddCat` as default.
*  Clear `NewTestCat` on `AddCategory()`. I had to add a property for   `NewTextCat` in `CategoryConfigViewModel` to do this, I'm not sure   if that's the cleanest way.
*  Don't add the `Category` to `Categories` if it already exists and is   enabled. This fixes a bug that you could have duplicate categories   listed.
*  If the category does already exist but is not enabled, enable it. I   had to make `Category` an `ObservableObject` for this to work, it   needed to be able to notify that it had been updated.
*  Removed default constructor for `Category` and setter for `Name`.   This prevents changing the `Name` of a `Category`, ensuring there   are never duplicates in `Categories`.

Now you can quickly create categories by typing in their name and pressing enter. If the text box is not focused, the OK button of the categories window will be submitted instead.

(This is the first time I've written C# and XAML since almost 8 years ago. Please let me know if I wrote any horribly unidiomatic code or anything.)

### DIFF
--- a/source/PlayniteUI/ViewModels/CategoryConfigViewModel.cs
+++ b/source/PlayniteUI/ViewModels/CategoryConfigViewModel.cs
@@ -13,22 +13,25 @@ namespace PlayniteUI.ViewModels
 {
     public class CategoryConfigViewModel : ObservableObject
     {
-        public class Category
+        public class Category : ObservableObject
         {
+            private bool? enabled;
             public bool? Enabled
             {
-                get; set;
+                get
+                {
+                    return enabled;
+                }
+
+                set
+                {
+                    enabled = value;
+                    OnPropertyChanged();
+                }
             }
 
-            public string Name
-            {
-                get; set;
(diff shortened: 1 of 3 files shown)


bburky:
I really wonder if `Categories` should really be something like a dictionary or a set, but I think that would be a separate change.

Removing the ability to rename `Category` objects isn't perfect, but it at least makes the current behavior not broken. It will at least force someone to think through it if they implement actually renaming categories. They would probably need to remember their old name to be able to find their old category name and replace it with the new one throughout the entire library.

Issue #919, opened 2019-01-06
Scripting API suggestions


bburky:
* Add an easy way to get a path to the extension directory. It's not too unreasonable for users to rename the extension folder's name, so manually deriving this could be error prone. (`__file__` should do this, but it evaluates to "\<module>" in IronPython, not a path to it's own location. But that's a separate IronPython bug I think.)

* Maybe allow easier configuration of script extensions. I need to prompt the user for the location of their Steam userdata folder and save that location somewhere. (It looks like `PlayniteApi.GetPluginConfiguration<TConfig>(IPlugin)` is actually accessible in IronPython. Maybe just create an IPlugin instance for script extensions and reuse it?)

* A REPL with access to the PlayniteApi would be really nice. It would be nice to incrementally test things against the API.

bburky:
Another:

Make `PlayniteApi.ExpandGameVariables(Game, GameAction)` also expand an emulator's Executable, WorkingDirectory and Arguments into the new GameAction, correctly handling OverrideDefaultArgs and AdditionalArguments.

Currently I have to manually lookup the emulator. Look up the Profile (using Linq in IronPython, which is weird but works). Manually expand each variable from the profile with  `PlayniteApi.ExpandGameVariables(Game, string)`. And append AdditionalArguments myself. I gave up for now.

PR #924, opened 2019-01-10
PowerShell and IronPython scripting improvements

### LINKED ISSUE #916: Add directory containing IronPython scripts to SearchPaths
Please clone this simple test extension to test IronPython imports:
https://gist.github.com/bburky/a213d97b81c1431ee9791e5831890af7

Currently importing a module in IronPython fails:

```
ImportError: No module named b.
```

After some debugging and reading StackOverflow, I discovered you need to manually add the directory containing the script to the ScriptEngine's SearchPaths. I almost wrote you a pull request but, `engine` is private to IronPythonRuntime and I didn't know how you wanted to expose configuring SearchPaths. I think where you _want_ to configure it is IronPythonScript's constructor.

As a quick hack I tested by modifying IronPythonRuntime like this and the test extension above correctly runs:

```
public void ExecuteFile(string path)
{
    var paths = engine.GetSearchPaths();
    paths.Add(Path.GetDirectoryName(path));
    engine.SetSearchPaths(paths);
    engine.ExecuteFile(path, scope);
}
```

### LINKED ISSUE #917: Allow optionally configuring the IronPython engine in debug mode
It is useful to run IronPython in debug mode while developing extensions. Running IronPython in debug mode allows setting breakpoints, single stepping and viewing Python local variables. (Sadly, no repl though.)

Running in debug mode does slow down IronPython significantly. And as far as I can tell it requires Visual Studio to do anything with, so this isn't anything most users need. Maybe only enable it when `Debugger.IsAttached` or something like only in debug builds?

You can enable debug mode like this:

```
public IronPythonRuntime()
{
    Dictionary<string, object> options = new Dictionary<string, object>();
    options["Debug"] = true;
    engine = Python.CreateEngine(options);
[...]
```

It may be worth documenting this too for people developing extensions.

bburky:
## Set IronPython engine to Debug mode when Debugger.IsAttached 

Resolves #917 

## Set current directory to location of IronPython scripts before executing 

Because the current directory "." is in the IronPython search path, this allows module imports to work.

Resolves #916

## Overhaul IronPythonRuntime

Avoid using `engine.Execute("...", scope)` during initialization, use C# APIs instead when possible.

Import Playnite.SDK.Models assembly using reflection instead of by DLL name.

Import Python files as modules using ImportModule instead of using ExecuteFile. Note, modules create their own new scope representing the contents of the module. To allow setting variables, set the scope to that of the newly imported module (discarding any previous scope or imported python file's variables).

Running Python files as modules improves their functionality. Now the special Python variables `__name__` and `__file__` are correctly set. Python extensions can now easily get a reference to the extension directory. Addresses part of #919.

Because Python modules run in a new empty scope, they won't be able to access any existing variables set in an existing scope. Preserve compatibility with existing Playnite extensions by instead injecting variables in to Python's `__builtins__` module which create global variables accessible in any Python module.

As a breaking change in a future version of Playnite, consider not creating global variables for `PlayniteApi`, `__logger` and all types in Playnite.SDK.Models. Instead add variables to `engine.Runtime.Globals`. These can still be accessed in Python, but only after deliberately importing them:

    import PlayniteApi
    from Playnite.SDK.Models import *
    
    PlayniteAPI.Dialogs.ShowMessage("foo")

The biggest risk to setting variables in `__builtins__` is that they appear in all Python modules, which may be third party modules used by the extension. This could have side effects if the module didn't expect the variable to exist. As a breaking change I would recommend moving the variables from `__builtins__` to  `engine.Runtime.Globals`.

Lookup functions in GetFunctionExits by variable name using `scope.GetVariableNames()`.

Modify GetVariable to return a dynamic variable which is called from IronPythonScript. InvokeExportedFunction may now throw RuntimeBinderException errors, but these are caught by the existing error handling. This may occur if InvokeExportedFunction is just a variable instead of a callable Python function. The user would see a dialog with the error:

> Failed to execute extension function.
> Cannot invoke a non-delegate type


bburky:
I tried to preserve compatibility with the existing extensions, all the global variables should behave the same. The only change is that Python extensions now have the current directory set to the extension directory. (Actually technically parent of the python file, which could be deeper if the .py file is nested inside a directory inside the extension folder.) I think this is actually a useful current directory to run extensions from, if they want the Playnite directory they can get it from `PlayniteAPI.Paths`. I'm not sure where PowerShell extensions run from, but you may want to make them consistent.

Running Python files as modules makes them behave far more like standard code is run. I managed to make the global Playnite variables still work by injecting them into `__builtins__`, but that's slightly hacky. I would rather make people import them themselves, I think it would be more Pythonic.

JosefNemec:
I haven't checked the changes yet, but I will probably merge this only into 5.0 branch because that will come will number of other breaking changes to extension API. So overall if you have good ideas how to improve things, don't look at backwards compatibility that much for that reason.

bburky:
Ok, I'll add an additional commit with the breaking changes then.

bburky:
One other thing is that Python modules can be a directory, not just a single file. My changes above should support multi file modules, but currently there is no way to specify these in the YAML because it detects extension types by filetype, which directories don't have.

I don't know enough PowerShell, but I suspect it has multi file modules too? Perhaps add an additional key in the YAML like `language: Python`?

bburky:
Okay, here's the version that has a breaking change in how variables are used in Python scripts. I had to modify `PlayniteScript` to take a list of variables to inject before even running the script file. I had to do this because a `import PlayniteApi` was running before the variable was injected.

I modified any exiting Python scripts to be compatible with the new Api. (Does FullPythonScript work? I couldn't figure out how to make all the hooks run.)

This needs testing of the PowerShell code, I haven't tested it at all. And probably a documentation update for the Python change.

bburky:
`on_script_loaded` appears to be unimplemented and `on_application_started` is undocumented here:
https://playnite.link/docs/tutorials/scripts/scriptingEvents.html. Should `on_script_loaded` exist? I would expect `on_script_loaded` to run when `on_application_started`, but also when reloading scripts. (Though you could just put code loose in the script, outside of functions, that will be run while loading the script itself.)

bburky:
Ok, I think I'm done. Please ask if you need any explanation or disagree with any changes I made.

JosefNemec:
Thanks, I'll review this when integrating 5.0 SDK changes (sometime in February).

bburky:
Setting the name of the PowerShell Runspace allows very easy debugging of extensions. Here's a quick demo, run the following commands inside PowerShell ISE: 

```
PS C:\Users\Blake> Enter-PSHostProcess -Name PlayniteUI
[Process:8168]: PS C:\Users\Blake> Get-Runspace

 Id Name            ComputerName    Type          State         Availability
 -- ----            ------------    ----          -----         ------------
  1 Library Expo... localhost       Local         Opened        Available
  2 RemoteHost      localhost       Local         Opened        Busy

[Process:8168]: PS C:\Users\Blake> Debug-Runspace -Name "Library Exporter"
Debugging Runspace: Library Exporter
To end the debugging session type the 'Detach' command at the debugger prompt, or type 'Ctrl+C' otherwise.
```

Then in Playnite, click Extensions → Export Library. The current extension's code should appear in the top pane. You get access to a REPL with access to all of the PlayniteApi!

```
[DBG]: [Process:8168]: [Library Exporter]: PS C:\Users\Blake\Code\Playnite\source\PlayniteUI\bin\x86\Debug\Extensions\LibraryExporter>> $PlayniteApi

Dialogs         : PlayniteUI.DialogsFactory
Database        : Playnite.API.DatabaseAPI
MainView        : PlayniteUI.API.MainViewAPI
Paths           : Playnite.API.PlaynitePathsAPI
ApplicationInfo : Playnite.API.PlayniteInfoAPI
WebViews        : PlayniteUI.WebView.WebViewFactory
Resources       : Playnite.SDK.DefaultResourceProvider
Notifications   : Playnite.API.NotificationsAPI
```

The above commands `Enter-PSHostProcess ...` and `Debug-Runspace ...` can also be run in VS Code's integrated PowerShell console to access a full debugger with a console, breakpoints, etc. Using VS Code's built in "PowerShell Attach to Host Process" doesn't work well because it seems to just select the first Runspace arbitrarily, and there may be multiple present if there are multiple PowerShell extensions loaded into Playnite each with their own Runspace.

JosefNemec:
I've pushed refactoring changes for 5.0 into `devel`, could you please re-base it on that?

bburky:
I added `Execute()` back to the scripting runtimes for testing.

`FunctionExecuteTest` and `GetFunctionExitsTest` are failing for PowerShell because PowerShell's `GetFunctionExits` expects the functions to be defined in a module, not defined in the runspace which is what `Execute()` does.

JosefNemec:
About failing tests, can you please update them if they are failing? I don't really feel comfortable merging changes with non-functional tests.

Also could you please update necessary doc files to reflect changes you made?

As far as `language` property in extension manifest it is better way then doing it by file extenions, feel free to add it and change how script extensions are loaded. PowerShell modules can contain more files, but loading is all defined/handled in module's manifest. You only specify that single manifest file when calling `Import-Module`.

Also I'm very sorry it took so long to review this.

JosefNemec:
Any update on this?

JosefNemec:
Just a note about what's going to happen with this since bburky is not responding. I definitely want to have script extensions implemented as modules, so it's getting integrated (or reworked in the same manner as this PR). However I (or somebody else if volunteer is found) should probably do it together with rework from IronPython to pythonnet #1206. Just to make sure we don't have two breaking changes after each other.

JosefNemec:
Closing since the work is continuing in #2040

Issue #899, opened 2018-12-25
Show exact zoom value on top of the Zoom slider


ClearStrelok:
Maybe something like this ? 

![slider](https://user-images.githubusercontent.com/38838369/50423158-21dd4300-0852-11e9-98e7-81f598fcb89f.png)

Mates1500:
is this being worked on? seems simple enough

ClearStrelok:
That was more of a suggestion, as I'm not sure how much interest there is in it, but I could try and propose it in a commit (once I semi-know what I'm doing that is).

JosefNemec:
I'm not sure how useful this information would be. The only value we can currently display is target pixel width of cover image (which is what the slider changes) and even that will not be accurate when non-100% DPI scaling is used.

JosefNemec:
I guess we could treat some specific value as being 100% zoom and then offer values that would represent zoom" from 50% to 200%, and only display calculated percentage while still handling pixel width in background.

Mates1500:
Yeah, that's what I meant. I could get to work on this by the end of the next week if you're not already working on it.

JosefNemec:
I don't plan to work on this any time soon so it's up for the grabs.

JosefNemec:
Released in 4.71

PR #947, opened 2019-01-27
Add percentage popup to the zoom slider upon MouseOver

Mates1500:
Implementation of the proposal specified in #899.
image

Took me a while to figure this one out due to my unfamiliarity with WPF, but this should do.

### DIFF
--- a/source/Playnite/Settings/ViewProperties.cs
+++ b/source/Playnite/Settings/ViewProperties.cs
@@ -123,9 +123,9 @@ namespace Playnite
             }
         }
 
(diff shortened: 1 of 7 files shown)


JosefNemec:
Looks goods. Two things:

- Slider goes for me from 56% to 144%. Can you make it 50-150?

- I don't particularly like how the popup follows mouse. It doesn't look smooth to me and sometimes it gets stuck and doesn't follow it unless I move mouse away from slider and back to it. It would be probably better to have it static.

Mates1500:
Alright, done

PR #959, opened 2019-02-02
update and fixed Simplified Chinese localization

L1cardo:


JosefNemec:
I found out after merging this that your localization file cannot be loaded, there's a duplicate for `LOCSettingsSteamLibrarySource` key. I fixed it, but I would appreciate if you could try if changes work properly before creating new PR in future. Thank you

PR #964, opened 2019-02-05
Update ScummVMImporter to use the refactored Game api

bburky:
The `State` property no longer exists on Game objects as of 3371c606. Change the extension to use the new `IsInstalled` setter.


bburky:
I was installing my copy of Sam & Max Hit the Road from when I was a kid. (Which I apparently backed up as BIN/CUE because it had CD audio, and was now tedious to mount. And I had to dig up a tool to extract the file from HFS volume because I had the Mac version. Fun. Fixing Playnite was the easy bit.)

bburky:
@josefnamec Also, the download link on the wiki is dead, the file has moved in the repo: https://github.com/JosefNemec/Playnite/wiki/Importing-ScummVM-games

JosefNemec:
Thanks for letting me know, I've update the wiki page.

Issue #989, opened 2019-02-09
Add option to show number of items when grouping


stnzbass:
It would be great if we could see how many games there's in each group in list and details view. 
Displaying or not the count could be toggle in the settings.
It would display the number of items between parentheses next to the group name.

stnzbass:
I could work on this.
I never worked with .net but since this is a pretty simple change I think I could do it.

JosefNemec:
Released in 4.73

PR #996, opened 2019-02-10
Add option to show grouped items count (#989)

stnzbass:
This is a PR for the issue #989 . It adds an option in the appearance settings to show the item count next to the group name in list or grid view.

This is my first time contributing on github and my first time working in .net (I usually work in java). I'm sorry if I made stupid mistakes.

I was not sure from what branch to branch from. At first I tried with devel but I could not get the application running (xaml parsing error) so I branched from master.

Also, I added a new language string in English and French. I'm not sure what to do with the other languages.

Thanks
JF

### DIFF
--- a/source/Playnite/Settings/PlayniteSettings.cs
+++ b/source/Playnite/Settings/PlayniteSettings.cs
@@ -265,6 +265,21 @@ namespace Playnite.Settings
             }
         }
 
+        private bool showGroupCount = false;
+        public bool ShowGroupCount
+        {
+            get
+            {
+                return showGroupCount;
+            }
+
+            set
+            {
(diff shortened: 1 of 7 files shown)


JosefNemec:
Thanks for contribution. Basing on `master` was a right choice, `devel` is currently in very unstable state.

The only thing I would change is the description of new setting option. Something like "Show item count on group descriptions" would probably describe more clearly what the option does since it also affects `Details` view and will also be applied to `Covers` view once we add proper grouping there.

stnzbass:
I had a hard time finding a concise description. English is not my first language. I say that but I also had a hard time for french... :)

PR #1019, opened 2019-02-15
Import LastPlayed information from Steam

bburky:
LastActivity data is imported from \steam\userdata\{steam_id}\config\localconfig.vdf

A button has been added to `SteamLibrarySettings` to trigger import. Updating a game's `LastActivity` if it skipped if is Playnite's value is newer than Steam's data.

I named things "LastActivity" throughout the code even though the Steam and the Playnite UI says "Last Played" usually. It seemed to match the code better, let me know if it's confusing.

I'm not sure if I have any Steam mods in my account, I couldn't test whether `localconfig.vdf` has underscores in AppIds like `sharedconfig.vdf`.

I added a `GetGamesLastActivityTest()`,  but I don't know how to actually get tests to run in Visual Studio. Could you please help?

I introduced multiple new localization strings. I added them to `english.xaml`, I'm not sure that needs to be done though?

`LastActivity` is now imported during `GetLibraryGames(...)` too. One caveat though: if a public library is being imported but `settings.IdSource == SteamIdSource.Name`, we won't have an API key, so we can't convert the `AccountName` to an `AccountId`. In that case I just skip it. Also if during import there are any errors accessing `localconfig.vdf`, I just log and ignore it. (It's possible a user is importing a library by username and they have never logged into that Steam account locally.) Errors are shown when manually clicking the import button though.

If "Force sync play time" is set, `LastActivity` will be updated every import. (I wonder though, should we [only update `LastActivity` if it is older than the current value](https://github.com/JosefNemec/Playnite/blob/master/source/Playnite/Database/GameLibrary.cs#L127)? And maybe only update `Playtime` if it is lesser too? Someone may have played the game in Playnite without launching Steam and the times could be out of sync.)

### DIFF
--- a/source/Playnite/Localization/LocSource.xaml
+++ b/source/Playnite/Localization/LocSource.xaml
@@ -276,8 +276,9 @@ Prefer data from official Store and where not available use IGBD.com database.</
     <sys:String x:Key="LOCSettingsWhatsSteamName">What's my account name?</sys:String>
     <sys:String x:Key="LOCSettingsLoginStatus">Login status:</sys:String>    
     <sys:String x:Key="LOCSettingsWindowTitle">Playnite Settings</sys:String>
-    <sys:String x:Key="LOCSettingsSteamImportCategories">Import Steam Categories:</sys:String>
-    <sys:String x:Key="LOCSettingsSteamImportCatFrom">from account:</sys:String>
+    <sys:String x:Key="LOCSettingsSteamImportMetadata">Import metadata from Steam account:</sys:String>
+    <sys:String x:Key="LOCImportSteamCategoriesLabel">Import Steam Categories</sys:String>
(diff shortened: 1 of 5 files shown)


bburky:
Ok, rebased on devel.

The more I look at this, the less I think a button in the settings dialog is a good idea though. I think most people who use this feature will use "Force sync play time" which causes this to happen automatically. The only issues that this _does_ interact with that drop down in the settings dialog: it has to know _which_ account to pull play time from. If you have any critiques on the settings dialog change, please let me know.

steam

JosefNemec:
You could alternatively add it as an option into "Extensions" menu. But then you would have to add some selection window for account as well.

bburky:
Eh, I think just having a button in settings is the best option then. It's the simplest UI I think.

PR #1037, opened 2019-02-22
Support building with VS2017 Enterprise (#1036)

### LINKED ISSUE #1036: Support building with VS2017 Enterprise
`build.ps1` hard-codes `"C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\MSBuild.exe"`. This can be auto-detected using [vswhere](https://github.com/Microsoft/vswhere).

This has the additional benefit of not relying on an English localization for the Program Files directory for a minimum build.

tylerszabo:
Fixes #1036

### DIFF
--- a/build/build.ps1
+++ b/build/build.ps1
@@ -38,6 +38,8 @@
     [string]$TempDir = (Join-Path $env:TEMP "PlayniteBuild")
 )
 
+$VSWHERE_CMD = "vswhere"
+
 $ErrorActionPreference = "Stop"
 & .\common.ps1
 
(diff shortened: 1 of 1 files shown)


tylerszabo:
@JanStefanski this one is blocking me from building directly off of master; if it's possible to expedite review I'd really appreciate it :)

build/build.ps1
```diff
@@ -150,7 +152,17 @@ if (!$SkipBuild)
...
 
     $solutionDir = Join-Path $pwd "..\source"
-    $msbuildPath = "c:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\MSBuild.exe";
+
+    if (-not (Get-Command -Name $VSWHERE_CMD -Type Application -ErrorAction Ignore))
+    {
+        $VSWHERE_CMD = (Join-Path $solutionDir "packages\vswhere.2.6.7\tools\vswhere.exe")
+        if (-not (Get-Command -Name $VSWHERE_CMD -Type Application -ErrorAction Ignore))
+        {
+            & nuget install vswhere -Version 2.6.7 -SolutionDirectory $solutionDir
```
JosefNemec: Doesn't work on my machine, please use `StartAndWait` instead.

tylerszabo: Can do.

So you have more information on how it fails so I can make my scripts more robust in general (for other projects).

tylerszabo:
I'm hoping this update works on your system; I'm wondering if the issue might have been not invoking `nuget.exe` explicitly (perhaps a different `nuget` wins on your `PATH`), so I used the same path as was used above in:

```powershell
    $nugetProc = Start-Process "nuget.exe" "restore ..\source\Playnite.sln" -PassThru -NoNewWindow
    $handle = $nugetProc.Handle
    $nugetProc.WaitForExit()
```

I still use `&` invocation below:

```powershell
    $msbuildpath = & $VSWHERE_CMD -version "[15.0,16.0)" -requires Microsoft.Component.MSBuild -find "MSBuild\**\Bin\MSBuild.exe"
```

Because I need to capture the output of `vswhere`.

build/build.ps1
JosefNemec: It says that `nuget` is not known command. I generally recommend using standard commands instead of expression invoke, or at least add obvious strings to quotes.

PR #1046, opened 2019-02-24
Use HTTPS for Steam links.

psychonic:
Most or all of these links already get redirected to HTTPS. This way we're saving that step. ~~Also fixed forum link going to hub instead of discussion forum.~~ (see comments)


JosefNemec:
Forum link pointing to the hub was actually intentional, but I agree that it was probably wrong labeling it `Forum`.

psychonic:
No worries. Reverted that part now.

Issue #397, opened 2018-01-05
Improve game description formatting


PhrozenByte:
The fact that Playnite expects game descriptions to be HTML caused some minor confusion in the past (#349); even though there's a informative tooltip now, it might still cause confusion - especially because I think that most users don't know what HTML is. Nevertheless is supporting HTML important because some stores use HTML for their game descriptions.

I suggest that the game description field supports Markdown. I guess everyone here on GitHub knows what Markdown is, so I think that I don't have to explain it. The biggest advantage of Markdown is that it is very intuitive (most importantly: paragraphs work as expected) and its simplicity - and, to return to the above "We must support HTML" requirement, it doesn't interfere with HTML in any way and it simply passes HTML through without any change. Just to emphasis this, Markdown parsers don't interpret Markdown inside HTML block elements - thus some `*` somewhere in a game description from Steam won't be parsed as Markdown. Consequently, adding Markdown-support doesn't break backwards compatibility.

I suggest using a [CommonMark](http://commonmark.org/)-compatible parser. There are some C# Markdown parsers out there Playnite could use (e.g. [`lunet-io/markdig`](https://github.com/lunet-io/markdig) or [`Knagis/CommonMark.NET`](https://github.com/Knagis/CommonMark.NET)).

JosefNemec:
I would actually don't mind getting rid of HTML support (yes we would have to convert Steam and GOG descriptions) in favor of something that doesn't render into HTML but FlowDocument instead (like [Markdown.XAML](https://github.com/theunrepentantgeek/Markdown.XAML)).

It would be faster and provide more unified look with the rest of application. Currently HTML descriptions are rendered by [HtmlRenderer.WPF](https://www.nuget.org/packages/HtmlRenderer.WPF), which is basically no longer maintained and has several issues with how the output looks.

nathanmerrill:
+1 to this from me.  Images in descriptions from Epic games use markdown format, which means that Playnite doesn't render them.

Deide:
Can CefSharp not render HTML?

Bluscream:
> 
> 
> Can CefSharp not render HTML?

that would allow the html to do things like run javascript. which you certainly wouldn't want

Bluscream:
If you don't think this would be abused: https://steamdb.info/app/737430/

it is just waiting for steams parser to fuck up in some way

![](https://i.imgur.com/gMWo3af.png)

JosefNemec:
[...]

PR #1071, opened 2019-03-13
Discord Store Library Extension

### LINKED ISSUE #854: Discord Store library support

bburky:
Adding installed games and importing metadata is fully implemented. Login and adding uninstalled games is unfinished.

Implementing uninstalled games requires finishing the login process. To finish the login process I need to extract the JWT `token` value from Javascript's  `localStorage`. There doesn't seem to be a built in API in CefSharp to do this, anything I can find online suggests [injecting a replacement `localStorage` object into `window`](https://stackoverflow.com/questions/35983961/how-can-i-read-edit-cefsharp-local-storage). Currently you hide all the internal details of CefSharp in the `IWebView` interface, could we maybe add some local storage functions on `WebView` objects? Do you want that in a seperate pull request?

Alternately I could intercept the XHR response to `https://discordapp.com/api/v6/auth/login` which contains the token, but that sounds equally complicated.

To display the description text from the Discord Store I need to format Markdown into HTML. Do you want me to add a Markdown library as a dependency, or just skip description for now. I know you have an issue #397 considering supporting Markdown in descriptions, I don't know if you want to do anything for that now.

I don't have Discord Nitro, I don't know how this will behave with it. Do Nitro games appear in the user's library or not? Do people want them listed in Playnite if they haven't ever downloaded them?

Let me know if you have any comments on is pull request. I tried to follow the design of the other library extensions.

Resolves #854

## Unfinished items

- [x] **Find out how to launch install dialog** `discord:///library/{Game.GameId}/uninstall` exists, but I can't find anything similar for installation. (I just launch the library `discord:///library/` instead and the user has to pick the game to install)
- [x] **Login**
   - [x] ~Inject a proxy JS object to read `localStorage` from C#~ Added EvaluateScript to IWebView.
   - [x] Extract JWT authentication `token` from `localStorage`
   - [x] ~Logout user before logging in (clear `localStorage`)~ "Clear web cache" should work.
- [x] **Import uninstalled games** (requires login)
   - [x] Fetch user's game list `https://discordapp.com/api/v6/users/@me/library`
   - [x]  Fetch last activity information `https://discordapp.com/api/v6/users/@me/activities/statistics/applications`
- ~[x] **Improve `FeatureType` and `Genre` enums**~ Removed
    - ~[x] Make JSON deserialization gracefully fail if unrecognized values are present?~
    - ~[x] Use `Description`, not the enum field name, when generating game metadata~
- [x] **Format store description Markdown into HTML** Formatted using MarkdownDeep.Net
- [ ] **Add tests**
- [ ] **Test with Discord Nitro**


JosefNemec:
As far as  localStorage and CefSharp goes, I think it would better to add generic JavaScript injection methods to `IWebView` rather then implementing something specific only for localStorage, if possible.

Descriptions: Add any dependency if you need one for converting MD to HTML. It will take some time until we implement #397, so don't wait for that.

`Genre` and `FeatureType` models: I don't really like storing the lists statically in a plugin. You would have to update it any time Discord decides to update the list, which would mean new version of a plugin (and currently new Playnite version). Metadata download should ideally download the meta list a parse it when needed.

I don't have a Nitro, I have idea how it works. I would simply go by what Discord API reports are part of you library. If it lists all Nitro games then import them, if there's a way how to distinguish between bought game and free from Nitro then maybe add an option to settings, for users to choose whether they want Nitro games imported or not (if they are not installed, we should import all installed).

bburky:
>Genre and FeatureType models: I don't really like storing the lists statically in a plugin.

Okay. I think we'll have to do without them entirely then. The only definitions of the enums are buried within files named things like `012d4cdfa848e4ebc4a2.js`, I don't think those files have stable URLs. And parsing JS would require some horrible regexes.

bburky:
This is the JS api I decided on: https://github.com/JosefNemec/Playnite/pull/1071/commits/6a73d2be127672e71c896bfe6a511ae60eb21311. I think it's fairly general purpose, other extensions may find it useful too. I can pull that out into a seperate pull request.

Here an example of using it, you can pass an anonymous function:
https://github.com/JosefNemec/Playnite/blob/716be4a420fe1a90a3d4142f3812db2f93985154/source/Plugins/DiscordLibrary/Services/DiscordAccountClient.cs#L42

JosefNemec:
Looks good to me, we can leave it as part of this PR in case you find something that has to be changed about it.

bburky:
I used [MarkdownDeep.Net](https://www.toptensoftware.com/markdowndeep/) from nuget to format the descriptions. I manually converted images to HTML because I had to format their internal `asset://` URLs into actual images, and had to center the images in the HTML.

One issue is that GIF images aren't too uncommon in the Discord store text and Playnite doesn't seem to play them.

bburky:
This is mostly done then except tests. Tests for other plugins look fairly short, what all do you expect?

Also, I wasn't sure how you wanted the API key stored, currently I'm just saving it in the extension settings. Other plugins seem to write it to a `tokens.json` file, is that the preferred option?

JosefNemec:
Doing unit tests for these plugins is tricky since you can't actually run them without client and games being installed. So in fact they don't classify as true unit tests :) The ones I have for other plugins were primarily used during development so I didn't have start the whole UI to test game detection capabilities. Basically if there's something you can write "true" unit test for then do it. Otherwise I won't be mad if there are no for these plugins :)

I store authentication tokens in separate file, because it's something that's not changeable by user, therefore it doesn't classify as something that should go into settings file.

JosefNemec:
I've pushed refactoring changes for 5.0 into `devel`, could you please re-base it on that? SDK changed and there will be merge conflicts.

bburky:
Rebase is mostly finished.

One real new bug that I don't quite know what's causing it: I'm getting this error "System.NotSupportedException: 'This type of CollectionView does not support changes to its SourceCollection from a thread different from the Dispatcher thread.'" thrown from `SelectableItem`. It prevents the metadata from working (....usually, sometimes it will work when the game being updated isn't currently selected I think).

Not sure if I completely got all the 5.0 api changes right, I mostly just fixed whatever errors were needed to get it to compile. Thank god for statically typed languages.

JosefNemec:
Could you attach call stack from when the `NotSupportedException` exception happens?

bburky:
Hmm, I'm not getting it to reproduce at the moment. I was getting it to trigger by causing an import using the batch Metadata Import dialog. I'll post it if I get a crash out of it again.

bburky:
Oh, I never moved the token to a tokens.json file. I'll do that soon.

JosefNemec:
I'm closing this since there doesn't seem to be any progress made on this, plus Discord closed their store recently.

PR #1085, opened 2019-03-21
Rework background update notifications

### LINKED ISSUE #995: Update notification steals focus
Original report: https://www.reddit.com/r/playnite/comments/aoy9lg/update_notifications_while_running_fullscreen/

Update notification should definitely not steal focus from any application currently in use. Ideally it should open in background or notification should  be delivered via Window's notification bar when Playnite window is not active.

Mates1500:
Fixes #995

The background update check now shows a notification in Playnite if Playnite is active, otherwise a balloon tip/w10 notification pops up in the background.

One thing that could perhaps improve this is binding the action of opening the update window simply by clicking the notification? I've noticed that there's an `InvokeAction` property in the `NotificationMessage` class. However, it does not seem to be utilized anywhere as of right now and every constructor call just passes in the null parameter.

### DIFF
--- a/source/PlayniteUI/App.xaml.cs
+++ b/source/PlayniteUI/App.xaml.cs
@@ -352,7 +352,7 @@ namespace PlayniteUI
             // Update and stats
             if (!PlayniteEnvironment.InOfflineMode)
             {
-                CheckUpdate();
+                NotifyIfUpdateIsAvailableDelayed();
                 SendUsageData();
             }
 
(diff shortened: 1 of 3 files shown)


JosefNemec:
Could you please re-base this on `devel`? I've pushed all refactoring changes for 5.0 and there will be merge conflicts.

The UI for notifications is not finalized, but that shouldn't be a blocker for this.

JosefNemec:
I forgot to mention that `InvokeAction` will be hooked up properly in final 5.0.

PR #1137, opened 2019-05-06
Fix devel build issues

Mates1500:
Was trying to do some work on #1085 but couldn't even run the program properly up until force removal of the `packages`, `bin` and `obj` folders, cleanup of the `.suo` file and a couple of VS restarts, then I hit these barriers.


JosefNemec:
I forgot that I had these projects unloaded and didn't see the build errors (we don't build them via build scripts atm.), however they are not required to build the Playnite itself, if you unload them in VS then application itself should build properly.

PR #1142, opened 2019-05-08
Rework background update notifications (devel rebase)

### LINKED ISSUE #995: Update notification steals focus

Mates1500:
Fixes #995 

So while working on this, I've noticed a minor annoyance with `NotifyIcon`, and that's if you don't `Dispose()` of it, the notification icon stays in the tray bar forever (or until you close Playnite). However, once I've hooked `Dispose()` to the notification close event, it would not stay in the W10 notification bar after a couple of seconds (so the notification would be utterly pointless while the person is let's say gaming in the background)

Therefore I've tried delving a little into `ToastNotification`, but quickly realized utilizing the API while the project doesn't target W10 specifically or is not a UWP app; becomes a real pain in the ass.

Finally, while traversing through the project, I've noticed that the `TaskBarIcon` class utilized for the tray operations of Playnite does provide the same balloon tip functionality without the above mentioned drawbacks. Well, except for one - hooking a custom click event into it doesn't seem so simple, so it can no longer be hooked into restoring the main window of Playnite upon clicking the notification.

### DIFF
--- a/source/Playnite.DesktopApp/DesktopApplication.cs
+++ b/source/Playnite.DesktopApp/DesktopApplication.cs
@@ -235,7 +235,14 @@ namespace Playnite.DesktopApp
         {
             var icon = new Icon(ThemeFile.GetFilePath("Images/applogo.ico", ThemeManager.DefaultTheme,
                 ThemeManager.CurrentTheme));
-            trayIcon.ShowBalloonTip(title, body, icon, true);
+            if (AppSettings.EnableTray)
+            {
+                trayIcon.ShowBalloonTip(title, body, icon, true);
+            }
+            else
+            {
+                WindowsNotifyIconManager.Notify(icon, title, body);
+            }
 
         }
     }
--- a/source/Playnite.DesktopApp/DesktopApplication.cs
+++ b/source/Playnite.DesktopApp/DesktopApplication.cs
@@ -230,5 +230,13 @@ namespace Playnite.DesktopApp
 
             return isFirstStart;
         }
+
+        public override void NotifyInWindows(string title, string body)
+        {
+            var icon = new Icon(ThemeFile.GetFilePath("Images/applogo.ico", ThemeManager.DefaultTheme,
+                ThemeManager.CurrentTheme));
+            trayIcon.ShowBalloonTip(title, body, icon, true);
+
+        }
     }
 }
--- a/source/Playnite/App/PlayniteApplication.cs
+++ b/source/Playnite/App/PlayniteApplication.cs
@@ -2,12 +2,10 @@
 using Playnite.Input;
 using Playnite.SDK;
 using Playnite.Plugins;
-using Playnite.Settings;
 using Playnite.ViewModels;
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Windows;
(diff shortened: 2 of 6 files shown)


source/Playnite.DesktopApp/DesktopApplication.cs
```diff
@@ -230,5 +230,13 @@ private bool ProcessStartupWizard()
 
             return isFirstStart;
         }
+
+        public override void NotifyInWindows(string title, string body)
+        {
+            var icon = new Icon(ThemeFile.GetFilePath("Images/applogo.ico", ThemeManager.DefaultTheme,
+                ThemeManager.CurrentTheme));
+            trayIcon.ShowBalloonTip(title, body, icon, true);
```
JosefNemec: This will throw null reference exception if tray icon is disabled in settings.

source/Playnite/App/PlayniteApplication.cs
```diff
@@ -442,15 +440,17 @@ public async void StartUpdateCheckerAsync()
...
-                                    new UpdateWindowFactory(),
-                                    new ResourceProvider(),
-                                    Dialogs);
-                                model.OpenView();
-                            });
+                                NotifyInWindows(updateTitle, updateBody);
+                            }
+                            else
+                            {
+                                Api.Notifications.Add(new NotificationMessage("UpdateAvailable", updateBody,
```
JosefNemec: I think we should add internal notification even if application is not active. In case that user dismissed the Windows notification and later opens UI he should still see that update is available.

JosefNemec:
What would be needed to make Windows notification open our update dialog when user clicks it?

source/Playnite.DesktopApp/DesktopApplication.cs
Mates1500: resolved in the recent commits below

source/Playnite/App/PlayniteApplication.cs
Mates1500: resolved in the recent commits below

source/Playnite/WindowsNotifyIconManager.cs
```diff
@@ -0,0 +1,24 @@
...
+        {
+            var notifyIcon = new NotifyIcon
+            {
+                Icon = icon,
+                BalloonTipTitle = title,
+                BalloonTipText = body,
+                Visible = true
+            };
+            notifyIcon.BalloonTipClicked += (o, ea) => { clickAction?.Invoke(); notifyIcon.Dispose(); };
+            notifyIcon.BalloonTipClosed += (o, ea) => { notifyIcon.Dispose(); };
```
Mates1500: This line I'm not exactly sure about. If we dispose of the icon, it properly disappears from the Windows taskbar. However, the W10 notification disappears along with it, so there's no chance the user will be able to see it after a gaming session.

If we don't dispose of the icon, it stays in the taskbar with no chance of cleaning it up later (even the click action stops working after the 5sec timeout)

Mates1500:
> What would be needed to make Windows notification open our update dialog when user clicks it?
I suppose getting `NotificationMessage.ActivationCommand` to start doing things in the UI should do the trick.

PR #1187, opened 2019-06-12
Determine MSBuild and NuGet locations consistently (#1186)

### LINKED ISSUE #1186: In build determine msbuild and nuget path consistently
There are a few places with similar and duplicate patterns for resolving `nuget.exe` and `msbuild.exe`. These checks can be moved to the build common to avoid either using different binaries or downloading unnecessarily.

tylerszabo:
Fixes #1186

### DIFF
--- a/build/build.ps1
+++ b/build/build.ps1
@@ -38,8 +38,6 @@
     [string]$TempDir = (Join-Path $env:TEMP "PlayniteBuild")
 )
 
-$VSWHERE_CMD = "vswhere"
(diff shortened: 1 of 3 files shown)


JosefNemec:
I've already done similar changes in `devel`, I think this might be no longer needed.

tylerszabo:
Indeed! Didn't realize `master` wasn't bleeding edge ;)

Probably best to discard and close both.

PR #1188, opened 2019-06-12
Fix crash when exception hit partway through metadata download

tylerszabo:
`downloaders` elements can have `null` values; need to null-check before calling `Dispose()`


tylerszabo:
This is not my day; I sync'd yesterday.

Issue #1140, opened 2019-05-07
Time/Last Played isn't imported from the origin launcher


ChildishGiant:
There are several similar issues but not any that are the same as this. Basically when importing games from origin, both time played and last played aren't imported. I know this is intended but this is basically a feature request

![image](https://user-images.githubusercontent.com/13716824/57327333-9d0b0b80-7106-11e9-8e3f-c604162bffb6.png)
![image](https://user-images.githubusercontent.com/13716824/57327420-cfb50400-7106-11e9-8ddb-fabec6fab089.png)

JosefNemec:
Play time import is currently only supported for Steam and GOG games: https://github.com/JosefNemec/Playnite/wiki/Feature-table-based-on-library-provider

tylerszabo:
I think I can implement this; from looking at the website Origin has a way to query playtime / last played metadata. I'm not sure how stable the API is, though. @JosefNemec would you rather wait for official support or is this something I should take a crack at?

JosefNemec:
Feel free to look at it, I'm not planning to implement it myself any time soon.

PR #1194, opened 2019-06-15
Sync PlayTime and LastPlayed from Origin #1140

tylerszabo:
Implements #1140

### DIFF
--- a/source/Plugins/OriginLibrary/Models/GameStoreDataResponse.cs
+++ b/source/Plugins/OriginLibrary/Models/GameStoreDataResponse.cs
@@ -23,11 +23,13 @@ namespace OriginLibrary.Models
         public class Platform
         {
             public string platform;
+            public string multiplayerId;
             public DateTime releaseDate;
         }
 
         public string offerId;
         public string offerType;
+        public string masterTitleId;
         public List<Platform> platforms;
         public string publisherFacetKey;
         public string developerFacetKey;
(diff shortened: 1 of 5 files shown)


vogler:
Is this supposed to work in Playnit 5.8?
For me it doesn't import times.
(It also doesn't import uninstalled games from Origin, but that's another issue.)

JosefNemec:
It only works if you have import of uninstalled games enabled and you are authenticated.

vogler:
image
Not for me.

JosefNemec:
Do you have some time already recorded for those games? Playnite will only download play time if none was recorded yet (if it's set to 0) or if "Force sync play time" is enabled.

vogler:
I didn't, but the issue somehow resolved itself. It now shows uninstalled and time played. 👍 
I guess the [feature table](https://github.com/JosefNemec/Playnite/wiki/Feature-table-based-on-library-provider) should be updated for Origin.

What's the reason for an option 'Force sync play time' and it not being the default? Shouldn't launching a game through Playnite always also update the time played recorded by Origin/others?

Another strange issue: 'Completion Status' only shows 'Played' for installed Origin games (not started) - for uninstalled games with 'Time Played' it's 'Not Played'. For Steam games it's the other way around ('Not Played' for installed, 'Played' for uninstalled).

JosefNemec:
Some games do not launch though official launcher. Mainly older EA games (Sim City 4, Mass Effect for example) will start directly without launching Origin first (it's not something we control, we just launch the executable and if Origin is required it will be opened automatically). That's why it's not enabled by default

Please open the completion status thing as different issue.

vogler:
Opened #1457.

Maybe it's not worth the effort, but if one would store the time from the server separately, one could check after playing if it changed, and if not add it to the local time. 'Time Played' would then always be the sum of server and local time.
Alternatively, is there a way of forcing a launch through the launcher (or maybe it's enough to have an option to always open the launcher before starting a game)?

Issue #303, opened 2017-12-17
Support for Bethesda and Epic Launcher


CantyCanadian:
I didn't see any tickets about it. Bethesda and Epic have their own launchers. While Epic's is more complex due to its support of Unreal Engine, it's still something that would be fun to have. Bethesda are getting more and more games in theirs so that would be great.

JosefNemec:
Epic already requested in #132 and #192

hammersyko:
I was also going to request the Bethesda launcher. Any word on this by any chance?

JosefNemec:
No update on this. It won't be done before #523 is implemented.

JosefNemec:
Support for installed games released in 4.59

RayRJJackson:
Any chance that we can authenticate with the Bethesda account without necessarily having to download the launcher?

ArtexJay:
> Any chance that we can authenticate with the Bethesda account without necessarily having to download the launcher?

Hope this can resolved sometime soon.

JosefNemec:
If somebody does the research and gives me description of APIs we can use to get user information then I'll implement it. Other then that I have no plans to work on Bethesda support in near future.

PR #1203, opened 2019-06-19
Implement user+pass authentication on Bethesda and Uplay accounts

RayRJJackson:
I dunno if this is the correct way to file a request, but...can you please add this methods of authentication to both Bethesda and Uplay libraries? Thanks.


JosefNemec:
Pull requests are for submitting new code changes. You need to use `Issues`. However both of these are already being tracked in #196 and #303

Issue #1213, opened 2019-06-25
KeePass use with library authentications keeps entering text in the main Playnite Window.


WolvenSpectre:
When reauthenticating my library authentications using the KeePass Password Vault instead of entering credentials in the Account Name and Password fields it goes back to the primary Playnite Window and types the info in the search box.

Normal behavior has the user go from the window with the fields for credentials and then to KeePass to find them and fill them out for you.

tylerszabo:
Playnite's `WebViewWindow` has no title, so KeePass can't detect which window to match for auto-fill. It looks like the `TitleChanged` event from `CefSharp.Wpf.ChromiumWebBrowser` could be passed through to change the window title. Appending " - Playnite" would make it appear like other web browser windows and should work for KeePass.

tylerszabo:
I'll take a crack at this.

JosefNemec:
`ChromiumWebBrowser` doesn't control window's title. You need to modify `WebView` class and change it though `window` field. The best solution would be to extend `IWebView` and allow plugin users to set custom title if they want to.

JosefNemec:
Scratch what I said in my previous comment, I haven't though it though :)

JosefNemec:
Done in 5.0

PR #1214, opened 2019-06-25
Show web page title in Web View Window titles #1213

tylerszabo:
Implements #1213


tylerszabo:
Here's what it looks like in action:

image

KeePass will now autofill with appropriate rules.

PR #1225, opened 2019-06-29
Wait script to finish before starting

umbertomarotta:
This is especially useful with Playnite library games


JosefNemec:
I'm afraid you'll have describe in more details why this change, and why only for `OnStarting` event and not others. I don't remember why we are doing async call here instead of synchronous one, but I think this needs thorough testing.

umbertomarotta:
It has been a few months now I'm using this edit in Playnite 4. I made a simple script that shows and image and plays a random sound 2 seconds before the game is actually launched. Thing is that some games, especially those added directly to Playnite, start even before the script releases control. Since OnStarted exists, one could think, as I did, that OnStarting executes before the games starts, making this event the only one to be time sensitive.

JosefNemec:
I don't think this is a safe change. We should change the api and offer both blocking and non-blocking (async) methods and let extension developer choose what's best for their scenario.

I've opened new issue #1257 for it.

Issue #1154, opened 2019-05-17
Entering text in the top bar's search field and clicking "clear filters" resets the filters but does not clear the field


wmassingham:
Playnite 4.74. "Clear filters" doesn't clear the text from the top bar, even though it does remove filters from the list. That button should clear both the filter panel and the top bar's field.

Reproduce:
1. Expand the filter panel
2. Enter text in the top search field (observe the same text filled into the left "name" field)
3. Click "clear filters"
4. Filter fields on left are cleared and the list is no longer filtered, but the top bar still has your text in it

T-0pel:
I could pick this up if no one else did yet.

T-0pel:
@JosefNemec I think this issue can be closed, as it was resolved in 5.0.

PR #1234, opened 2019-07-06
Filter text box synchronisation issue

T-0pel:
Related to #1154. When opening the filter panel there is a possibility of the two name text boxes becoming out of sync, after which they do not get cleared when pressing the Clear Filters button. The binding breaks when the "backspace" icon is pressed in either of the "name" text boxes. The boxes will still filter the games, but will no longer be in sync.

Steps to reproduce:

1. Open filters
2. Type in either of the "name" text boxes (at this point they are in sync)
syncWorking
3. Click on the "backspace" icon in either of the text boxes
4. Start typing to the other one - at this point they are out of sync and will not get cleared either
syncNotWorking

I have tracked the issue to the TextBox in the SearchBox class. Changing Text Binding from OneWay to TwoWay fixes this issue and both textboxes are in sync again.


T-0pel:
I have just realized that the problem with the "Clear Filters" button is not specific to the "name" textbox. In the current version if you type into any of the filter text boxes, clear them with the "backspace" icon, and then type something again, it will not get cleared by the "Clear Filters" button.

This PR fixes the issue for all the filter text boxes.

JosefNemec:
I don't remember exactly why, but there was a reason the binding was originally set to `OneWay`. It might not be needed anymore in 5.x, but this could potentially cause multiple calls to filter refresh. Can you check that that's not the case?

T-0pel:
Sure I can take look at it. The "OneWay" was added in [this commit](https://github.com/JosefNemec/Playnite/commit/285a15316364e481e44cf1e0a15b768b7c602bdc). So the possible issue is that the filter refresh will get called multiple times?
I will also take a closer look at the "backspace" button because it is weird that it stops working only after clicking at it.

T-0pel:
I found the issue with the "backspace" button, it looks like setting the Text property to empty string [does not actually trigger data bindings](https://stackoverflow.com/questions/16236392/how-to-clear-a-textbox-once-a-button-is-clicked-in-wpf#comment23225892_16236442). Using the Clear() method of the TextBox triggers data bindings and does not break the functionality of the text box.

I have modified the code, so take a look @JosefNemec and let me know what you think. I have changed the binding mode back to OneWay just to be safe, although I have been unable to replicate multiple calls to filter refresh with the TwoWay mode.

Issue #826, opened 2018-11-14
Automatically remove uninstalled games


JosefNemec:
Original request: https://www.reddit.com/r/playnite/comments/9wxauj/can_i_automatically_remove_uninstalled/

I don't think we should do this automatically, but add some menu function to do it on demand instead.

Pralom:
Maybe if it warns you that the directory for this game isn't found and a pop up appears asking you to remove it, it will be better.

kllr007:
Yeah this is annoying. Playnite now showing games that are not installed and it is confusing.

T-0pel:
@JosefNemec I could take a look at adding this as a menu function that can be launched manually. It would check whether the game .exe file exists. If it does not, it will remove the game from the library. I think it should only target games that were not added automatically by the plugins. Please let me know what you think.

JosefNemec:
@T-0pel The original request was for emulated games specifically, so just checking exe is not enough. You would have to make separate checks for exe for standard games and rom check for emulated ones, which shouldn't be a problem.

It should definitely be added as separate menu function and only executed on demand. Also we should show the list of games that will be removed and ask for configuration to make sure we don't remove something by accident. Maybe even add an option to deselect game from the removal.

And yes it should only work with games that were not added via library plugin (since they would be re-imported back anyways).

matdombrock:
I would love to see this implemented if possible.

ElTioRata:
This is annoying me, please implement this.

Ansem93:
Any hope to see this feature implemented? 
I think it should be an option to have uninstalled games automatically removed or at least a manual check. It's incredibly annoying to have the uninstalled game still in the home page

reloxx13:
+1

NineKain:
+1

OxygenEgg:
+1

Getroekah:
+1
This would save me so much headaches

apotenza92:
I think this should be considered as a user preference that's on by default. 
If something is uninstalled via any means - whether Playnite or directly via Steam etc. - Playnite should have a way to remove that game from the user's library automatically during the next game library update. 

I personally think this should be on by default as it's confusing for new users to see uninstalled games still visible by default in Playnite.

Pal1ndrome:
I uninstalled several games from Xbox and Steam, refreshed my Playnite library (also closed and reopened), and the games still show up. Tried with and without the "Installed" filter.

PR #1248, opened 2019-07-16
Uninstalled games not being tagged as uninstalled

T-0pel:
Possibly related to issue #826.

With the "Import installed games" setting the GameDatabase.ImportGames method only updated properties for games that were installed on the machine. Games that were uninstalled were still left tagged as installed, which could lead to errors when trying to launch a non-existent game.

This patch adds a check to make sure that games are now marked as uninstalled if they are no longer installed in the client. It also does a minor code cleanup and adds unit tests to test the new behavior of the import games method.


source/Playnite/Database/GameDatabase.cs
```diff
@@ -712,12 +711,13 @@ public Game ImportGame(GameMetadata metadata)
         public List<Game> ImportGames(LibraryPlugin library, bool forcePlayTimeSync)
         {
             var addedGames = new List<Game>();
-            foreach (var newGame in library.GetGames())
+            var libraryGames = library.GetGames().ToList();
```
JosefNemec: Why conversion to list? Isn't that unnecessary here?

source/Playnite/Database/GameDatabase.cs
```diff
@@ -753,7 +753,13 @@ public List<Game> ImportGames(LibraryPlugin library, bool forcePlayTimeSync)
                 }
             }
 
-            return addedGames;        
+            // Set the uninstalled property for games that were removed
+            foreach (var game in Games.Where(g => g.PluginId == library.Id && g.IsInstalled))
+            {
+                game.IsInstalled = libraryGames.Any(lg => lg.GameId == game.GameId && lg.IsInstalled);
```
JosefNemec: You need to call `Games.Update` otherwise the information won't be persistent between app launches. Also only call it when the actual `IsInstalled` state changes (ideally don't even do the assignment to prevent property change notification to UI).

source/Playnite/Database/GameDatabase.cs
T-0pel: I am using it to [prevent multiple enumeration](https://www.jetbrains.com/help/resharper/2019.1/PossibleMultipleEnumeration.html) as the library games IEnumerable is now used in the foreach and the isInstalled check.

source/Playnite/Database/GameDatabase.cs
T-0pel: Yes, that was a mistake on my part, sorry.

I have pushed a fix that changes the code to behave as you have described.

JosefNemec:
I will probably have to revert this, see #1288. Overall this might also cause some other issues we are not seeing right now. The import should probably modify only games that have been returned by the plugin and leave others alone, just to be sure.

T-0pel:
Hi, I am sorry that this is causing issues, I will investigate it tonight.

However, as you have mentioned in the issue itself, this seems to be pointing to a larger issue with installed games not being recognized. I will ask for more details and test it as well.

I understand if you want to remove this functionality, I just think that it is a bit confusing when an uninstalled game keeps its installed tag, and it can lead to errors when launching the game.

We could just set the tag on games that have not been manually edited (if there is a way to check that), or we could set the uninstalled tag when the game errors out upon start.

JosefNemec:
This issue should only occur if user disables import of uninstalled games. I think in that case it does makes sense that "uninstalled" state is not being updated.

T-0pel:
If I understand the code and the issue correctly, this will occur even with the import of uninstalled games setting, because an installed game is not returned as part of the GetInstalledGames method.

Therefore it will get added as an uninstalled game [in the *Library class](https://github.com/JosefNemec/Playnite/blob/286ded37f82e31702965f21618e87581e43c040f/source/Plugins/GogLibrary/GogLibrary.cs#L236) and marked as uninstalled [in the GameDatabase class](https://github.com/JosefNemec/Playnite/blob/286ded37f82e31702965f21618e87581e43c040f/source/Playnite/Database/GameDatabase.cs#L726).

T-0pel:
Issue #1288 was caused by corrupted game entries in the libraries and it disappeared after verifying the installation of the games.

I have been able to reproduce it with GOG, but not with Steam.
Steps to reproduce - GOG:

1. Copy folder with the game to a different location.
2. Uninstall game via the GOG client.
3. Import game from the copied folder via Manage installation -> Import folder.

Now the game is playable and marked as installed, but it will not show in the programs list. This issue would have happened before this PR (I checked in version 4.74) when the "Import uninstalled games" setting is true. Verify installation fixes of the game fixes it. As this is an issue with the library, I am not sure there is much that can be done to fix this problem.

I understand your point of view, the user does not expect that this setting will change the tags on games that are already in the library. However, I think there is a use case for setting the uninstalled tag on games that no longer work. As mentioned in #826, the functionality could be changed to an on-demand function or be presented as an option when the game is not successfully launched.

JosefNemec:
The issue with GOG games not being properly marked as installed if you just import the game is [a known issue](https://github.com/JosefNemec/Playnite/wiki/Known-Issues#some-gog-games-appearing-as-uninstalled-in-playnite) for some time now (the same also happens with some Battle.net games).

My main issue is that I'm not sure we should be changing install state for games that were not returned by the library plugin during import. Playnite has no idea why plugin returned only these games and it might be a bad idea to modify other games if we don't know why there were not included in import.

I also think that the core issue might in the fact the the import itself is not done by the plugin, but Playnite. We should maybe move that functionality to each plugin and let the plugin work it out based on its settings.

PR #1249, opened 2019-07-17
Replaces %CDN_HOST_MEDIA_SSL% for Steam games

### LINKED ISSUE #1205: %CDN_HOST_MEDIA_SSL% not replaced
image

%CDN_HOST_MEDIA_SSL% should be replaced with steamcdn-a.akamaihd.net but that's not being done on Hollow Knight's description.

image
Here ya go

KLee1248:
Adds a converter to automatically replace %CDN_HOST_MEDIA_SSL% with
steamcdn-a.akamaihd.net.

Affects Hollow Knight metadata import.

Fixes #1205.

Screenshot after fix:
FixScreenshot

### DIFF
--- a/source/Plugins/SteamLibrary/SteamMetadataProvider.cs
+++ b/source/Plugins/SteamLibrary/SteamMetadataProvider.cs
@@ -279,6 +279,11 @@ namespace SteamLibrary
             return metadata;
         }
 
+        internal string ParseDescription(string description)
+        {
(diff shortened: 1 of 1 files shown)


JosefNemec:
I think this should be done in `SteamMetadataProvider` where we assign the description. The original source data should stay unchanged in case we need to process it in some other place as well.

KLee1248:
Updated the PR to make the conversion in the `SteamMetadataProvider`.

Issue #1227, opened 2019-06-30
Make account linking settings more clear


JosefNemec:
Right now account linking is directly tied to import of uninstalled games. That can cause problems if user only enabled import of installed games, but still wants to get features available only when account is linked (like import of play time).

T-0pel:
I would like to take a look at this if no one else is working on it. I think it is a great idea, I was really confused by this behavior when first trying the application.
It could also go well with this issue #826 because uninstalled games are getting automatically removed when the import of uninstalled games is turned on, but not when it is off.

JosefNemec:
Sure, just couple notes:

- this will have to be done for every library plugin separately (since library settings are handled by plugin itself)
- even though each plugin handles it separately we should implement this in some unified manner
- I was thinking adding something like "Connect Account" button

#826 is specifically about actually removing the game from Playnite's library. We currently don't remove game if import of uninstalled games is enabled, we just update "Installed" status of the game.

T-0pel:
Cool, that is in line with what I was thinking. I will take a look at it and let you know when I have some PoC ready.

Regarding #826, maybe I misunderstood, but it looked to me that the original post was about games being classified as "Installed" even when they are uninstalled in the client:

> For example say I try out a PS2 rom and decide I don't like the game so I delete it, it's still in Playnite and is useless since it won't run and is **still classified as an installed game**, and I have to track it down and delete it manually.

I have tested it with GOG and when only the "Import installed games" option is selected, the games uninstalled via GOG are still marked as installed in Playnite, even after updating the game library. When the option "Import installed games" is selected, the uninstalled games get correctly marked as uninstalled after library update. That is why I thought there might be some relationship, but I have not looked into it yet.

T-0pel:
I thought about the best way to do this and I have one main question. Do you want the "Connect Account" and "Uninstalled Games" settings to be completely independent of each other?

If not, we could just make the "Uninstalled Games" a subsetting of the "Connect Account", in which case we could keep the logic of the current methods. When only the "Connect Account" is checked we would filter the library games for the installed ones and ignore uninstalled ones. With both settings checked the uninstalled games will get imported as well (the way it works now).

But if the settings should be independent of each other we would need to modify the current logic. In this case, uninstalled games would now need to get imported without account features (i.e. playtime), and account synchronization will get handled by a new method.

Please let me know what you think @JosefNemec, thanks.

JosefNemec:
[...]

PR #1258, opened 2019-07-25
Separate account linking settings from uninstalled games import

T-0pel:
Issue #1227.

Modified the settings for all plugin libraries supporting the uninstalled games import. Users can now synchronize their game stats without having to import uninstalled games. Added Czech translation of the button text, for other languages I would have to use Google Translate.

I am used to doing code cleanup on classes that I work with, but I do realize that it makes the PR harder to follow. Let me know if this is ok or you would prefer for me to leave the other parts of the class intact.

I wanted to add unit tests, but they would require at least some refactor and the PR is already large enough without it. I can add them as separate issue/PR if you don't mind.


T-0pel:
@JosefNemec Hi, did you get a chance to look at this PR? I can modify it or recreate it without the code cleanup if you want.

JosefNemec:
Sorry it took me so long to get to it, I'm not feeling very well lately. Anyways I only checked how it looks like in the application and I think it's ok.

Please remove the refactoring/cleaup changes since it makes it so much harder to review the code. You can open separate PR for it later.

Also don't make changes to localization files in repository, they will not get pushed to Crowding and will be overwritten once I do push from Crowdin to GitHub. Only original English source file should be updated. Translated string should be only updated via Crowdin.

T-0pel:
No worries, I hope you feel better soon @JosefNemec!

I removed the code cleanup, I only left fixed typos and removed unnecessary spaces. I removed the Czech localization as well. It is still a lot of changes because I had to change each supported plugin.

As another PR, It could help to create some parent class for plugins that allow account connection. Something like "LibraryPluginWithAccount" that inherits from "LibraryPlugin" which will implement the GetGames method with settings from a new settings class "LibraryWithAccountSettings". This will limit code duplication and it would also make it easier to write unit tests for this functionality. Please let me know if you think that makes sense, otherwise, I will take a look at some other issue.

JosefNemec:
Sorry for no response for such a long time. I will be merging this to Playnite 6 once I start working on it (very soon).

T-0pel:
Cool, I understand, I also did not have much time during the summer. Let me know if there is anything else you want to me to change. I will take a look at #826 this week.

JosefNemec:
I merged this manually into `devel6`.

Issue #1285, opened 2019-07-27
Splash screen cannot be turned off.


Demiko:
In newest version a splash screen was added and there's no way of turning it off. It's a highly intrusive behavior when Playnite launches on computer start and it should be at least togglable if not removed entirely as it was before.

craigt00:
@JosefNemec - do you mind me picking up this one?  My intention is to add a new checkbox option to 'Settings.../General' to control this (defaulted to checked to show the splash screen).

JosefNemec:
Splash screen was added because there can now be a noticeable delay before any Playnite window is displayed after staring our executable (due to new theming system). Splash screen gives instant feedback that Playnite is actually starting.

However I agree that it's probably not necessary to show splash screen if Playnite is starting automatically during Windows boot.

@craigt00 Sounds reasonable, it should be enabled by default however.

craigt00:
Hi @JosefNemec  That makes sense to me.  Since it's specifically the splash screen when Playnite runs at startup which is the problem, I won't add the extra option checkbox and always suppress the splash screen in that scenario.  Starting Playnite via the desktop will display the screen as it does now.

JosefNemec:
This was done in Playnite 5.8.

PR #1316, opened 2019-07-29
Suppress splash screen when Playnite starts at Windows startup

craigt00:
For #1285 

Added a new startup argument named 'hidesplashscreen' which suppresses the splash screen when Playnite starts.  This parameter is added to the Playnite.lnk file which is generated when a user selects to start Playnite at windows startup (via Options...), thereby preventing the splash screen from appearing at Windows startup but retains the splash screen if Playnite is started manually.

### DIFF
--- a/source/Playnite/Settings/PlayniteSettings.cs
+++ b/source/Playnite/Settings/PlayniteSettings.cs
@@ -1351,7 +1351,13 @@ namespace Playnite
             if (runOnBootup)
             {
                 FileSystem.DeleteFile(shortcutPath);
-                Programs.CreateShortcut(PlaynitePaths.DesktopExecutablePath, "--hidesplashscreen", "", shortcutPath);
+
+                var hideSplashScreenOption = new CmdLineOptions()
+                {
+                    HideSplashScreen = true
+                };
+
+                Programs.CreateShortcut(PlaynitePaths.DesktopExecutablePath, hideSplashScreenOption.ToString(), "", shortcutPath);
             }
             else
             {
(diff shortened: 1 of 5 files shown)


source/Playnite/Settings/PlayniteSettings.cs
```diff
@@ -1351,7 +1351,7 @@ public static void SetBootupStateRegistration(bool runOnBootup)
             if (runOnBootup)
             {
                 FileSystem.DeleteFile(shortcutPath);
-                Programs.CreateShortcut(PlaynitePaths.DesktopExecutablePath, "", "", shortcutPath);
+                Programs.CreateShortcut(PlaynitePaths.DesktopExecutablePath, "--hidesplashscreen", "", shortcutPath);
```
JosefNemec: Create new `CmdLineOptions` object with appropriate settings and then pass the result via `ToString` method instead for direct string.

craigt00: No worries.  Pull request updated.

kremuwakarz:
I've replaced explorer.exe for Playnite.FullscreenApp.exe in windows shell so Playnite can now launch much faster! this is better than "run playnite when computer starts" option however the splash screen is annoying so I need a real ini setting for disabling splash screen or at least paramater for shortcut like -hidesplashscreen (so I could use .bat file to run Playnite.

JosefNemec:
https://github.com/JosefNemec/Playnite/wiki/Cmdline-arguments

PR #1359, opened 2019-08-10
Bug fix:  Playnite displays an error during computer shutdown (#866)

### LINKED ISSUE #866: Application hangs during shutdown
I leave Playnite running in the background at all times, and when I shut down my PC, the shutdown doesn't happen as Playnite hangs and needs user input to continue. This happens every single shutdown.
img_20181211_153654

Additionally, it would be nice if the application would remember your last position between sessions, and if it could be launched minimized to tray, without needing to enable "minimize to tray."

craigt00:
Fixes #866 

As discussed, I've added a condition into Application_SessionEnding which stops Playnite from attempting to release it's unmanaged resources (including CefSharp) if the computer is shutting down.  The current user logging off will continue to clean up resources as usual and this doesn't trigger the bug to occur.

As a side note, I did attempt to upgrade to the latest stable version of CefSharp (73.1.130) but this didn't fix the issue.  Therefore, since I'm not clear on the wider impact of updating CefSharp I've reverted this change in this pull request.

### DIFF
--- a/source/Playnite.Common/Programs.cs
+++ b/source/Playnite.Common/Programs.cs
@@ -52,10 +52,10 @@ namespace Playnite.Common
     {
         private static ILogger logger = LogManager.GetLogger();
 
-        public static void CreateShortcut(string executablePath, string arguments, string iconPath, string shortuctPath)
+        public static void CreateShortcut(string executablePath, string arguments, string iconPath, string shortcutPath)
         {
             var shell = new IWshRuntimeLibrary.WshShell();
-            var link = (IWshRuntimeLibrary.IWshShortcut)shell.CreateShortcut(shortuctPath);
+            var link = (IWshRuntimeLibrary.IWshShortcut)shell.CreateShortcut(shortcutPath);
             link.TargetPath = executablePath;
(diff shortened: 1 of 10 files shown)


JosefNemec:
I don't like that you are completely skipping release of all resources. We should only skip CefSharp stuff if that's the issue. Everything else should still be released/disposed properly.

craigt00:
Agreed.  I've refactored the solution so the condtional is now in the CefTools class.  How do you feel about that?

JosefNemec:
I don't like this approach either (I don't think that use static property in this scenario is appropriate). Why not just add argument to `ReleaseResources` that would skip disposal of CefSharp? You can then replace `Quit` call in shutdown handler with that (and also call app shutdown since you would not be using Quit anymore).

JosefNemec:
I fixed it in 86e2ebbfdd2f34463a3ec1a06c50b2269a7f327a

craigt00:
Thanks @JosefNemec.  Apologies for the (very) late response.  My work IRL has consumed a lot of my time over the last few weeks.

PR #1382, opened 2019-08-18
Update to add new emulator profiles

darklinkpower:
Added bsnes, Mesen-S, melonDS and higan emulator profiles.

### DIFF
--- a/source/Playnite/Emulators/Definitions.yaml
+++ b/source/Playnite/Emulators/Definitions.yaml
@@ -1520,3 +1520,105 @@
       Platforms: [Adobe Flash]
       ImageExtensions: [swf]
       ExecutableLookup: ^flashplayer_\d+_sa\.exe$
+
+- Name: bsnes
+  Website: 'https://bsnes.byuu.org/'
+  Profiles:
+    - Name: Default
+      DefaultArguments: '"{ImagePath}"'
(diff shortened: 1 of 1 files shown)


source/Playnite/Emulators/Definitions.yaml
```diff
@@ -1520,3 +1520,105 @@
...
       ImageExtensions: [swf]
       ExecutableLookup: ^flashplayer_\d+_sa\.exe$
+
+- Name: bsnes
+  Website: 'https://bsnes.byuu.org/'
+  Profiles:
+    - Name: Default
+      DefaultArguments: '"{ImagePath}"'
+      Platforms: [Super Nintendo Entertainment System]
+      ImageExtensions: [sfc, smc, zip]
```
By6r9: From v108 7z is also supported.

By6r9:
nst
On my test Playnite `Nintendo Satellaview` is checked as it should be.

source/Playnite/Emulators/Definitions.yaml
darklinkpower: Thanks, fixed it.
And you are right, Satellaview is checked now. I wonder if I saw wrong or what happened but it's working now.

darklinkpower:
Added SameBoy and bsnes-hd profiles to the PR

PR #1417, opened 2019-09-15
Term Changes to Steam links and one more added

darklinkpower:
Proposed changes
`Forums` -> `Community Hub`
`Discussions` (I added this one)
`Store` -> `Store Page`
`Wiki` -> `PCGamingWiki`

### DIFF
--- a/source/Plugins/SteamLibrary/SteamMetadataProvider.cs
+++ b/source/Plugins/SteamLibrary/SteamMetadataProvider.cs
@@ -293,10 +293,11 @@ namespace SteamLibrary
                 Name = downloadedMetadata.ProductDetails?["common"]["name"]?.Value ?? downloadedMetadata.GameInfo.Name,
                 Links = new List<Link>()
                 {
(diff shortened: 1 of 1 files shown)


JosefNemec:
Did you make sure this is unified across all other plugins? It doesn't make sense to me to rename it only for Steam.

darklinkpower:
I checked all the other ones after I made the PR and to my surprise they were all named exactly PCGamingWiki already, it was just the steam plugin that wasn't. 

The only thing that wouldn't be unified would be the term `Store Page` because I'm unsure if it should be changed in all the other ones, if you are fine with it I can make the term change in the other ones too.

JosefNemec:
It should be unified across all of them. If it points to store page then why should we call it differently for different stores?

darklinkpower:
I made PR to the GOG (#1422), Origin (#1421) and Itchio (#1420 ) libraries to change the term "Store" to "Store Page", that way it will be unified.

PR #1443, opened 2019-09-21
Add Achievement links to Steam Games

darklinkpower:
The link opens in the Steam client, that way there won't be problems in case you are not logged in on Steam in the web browser. I figured this can be a solution until there's native achievement support in Playnite


JosefNemec:
I'm not sure I like that you are adding this even if the game doesn't support achievements (at least I don't see any check for it).

PR #1469, opened 2019-09-29
Add Galaxy 2.0, IGDB, New Steam UI, EGS and Twitch Aspect ratio presets for Grid View

darklinkpower:


JosefNemec:
Please do not separate functionality and related localization strings with different PRs, it doesn't make sense. All the changes necessary for the changes should be in the same PR.

JosefNemec:
Also I don't see a point of 1:1 preset, you can easily add that manually. The whole point of preset is for the ratios that can't be easily calculated or are known.

darklinkpower:
Got it, sorry I'm still figuring how Github works. I agree that it can easily be added manualy, although maybe I saw a point in adding it more for easiness than for it being difficult to figure out, but I deleted it just now, let me know if there's something else that should be fixed.

darklinkpower:
Found a small bug that prevents the Galaxy 2.0 and Epic Games Store aspect ratio (171:241) to be used at least manually; any value higher than 100 can't be used, it gets reverted to the previous value.

darklinkpower:
I modified the max value to 350 in both fields in case another aspect ratio gets added in the future, the highest number right now should  be 241 for the GOG2 preset, but while not added, the Origin aspect ratio is 231:326

JosefNemec:
Just two things:
- `MaxValue` on num boxes. Please change it to 999, so we don't have to rise it again in future :)
- I think we should not bundle together multiple store options. For example if in future GOG decides to change the format, then we would have to change  a lot of things as opposed to just adjusting aspect values.

darklinkpower:
Got it, increased values.
For point 2, do you mean separating presets by individual stores even if they share the same aspect ratio? If it's that, I agree, I modified the files to have them individually.

PR #1478, opened 2019-10-02
Localization strings for game links

darklinkpower:
To change terms for the different languages. It improves the user experience in Playnite in case you are using a language other than english,  because the current terms don't apply to the other languages.


source/Playnite/Localization/LocSource.xaml
```diff
@@ -762,21 +762,48 @@ Version {2}
...
+    <sys:String x:Key="LOCBattleNetLinksPCGW">PCGamingWiki</sys:String>
+
+    <!--Bethesda-->
+    <sys:String x:Key="LOCBethesdaLinksPCGW">PCGamingWiki</sys:String>
+
     <!--itch.io-->
     <sys:String x:Key="LOCItchioSignInNotif">itch.io client will now open. Please sign in and then close this message.</sys:String>
     <sys:String x:Key="LOCItchioSignInWaitMessage">Waiting for user to sign in, please close this when you're done…</sys:String>
     <sys:String x:Key="LOCItchioClientNotInstalledError">Itch client version 25 or newer is not installed.</sys:String>
+    <sys:String x:Key="LOCItchioLinksPCGW">PCGamingWiki</sys:String>
```
JosefNemec: We don't need localization string for PCGamingWiki, it's a website name, nobody is going to translate it. And we definitely don't need duplicated version of that string for every library.

source/Playnite/Localization/LocSource.xaml
```diff
@@ -762,21 +762,48 @@ Version {2}
...
 
     <!--GOG-->
     <sys:String x:Key="LOCGOGUseLogin">Account Login</sys:String>
     <sys:String x:Key="LOCGOGUseAccountName">Account Name</sys:String>
     <sys:String x:Key="LOCGOGAccountSyncDescription" xml:space="preserve">Login to your account or use your public account name to sync your full library.
 Note that the Account Name option only works for accounts whose game libraries have been set in GOG's privacy settings to be publicly visible.</sys:String>
+    <sys:String x:Key="LOCGOGLinksPCGW">PCGamingWiki</sys:String>
+    <sys:String x:Key="LOCGOGLinksForum">Forum</sys:String>
```
JosefNemec: Please don't create separate string for each library. "Forum" (and other strings like Store) is pretty universal and I don't see a point of creating separate variants of it for each library provider.

source/Plugins/BattleNetLibrary/BattleNetMetadataProvider.cs
```diff
@@ -27,7 +27,7 @@ public override GameMetadata GetMetadata(Game game)
                 Links = new List<Link>(product.Links)
             };
 
-            gameInfo.Links.Add(new Link("PCGamingWiki", @"http://pcgamingwiki.com/w/index.php?search=" + product.Name));
+            gameInfo.Links.Add(new Link("{DynamicResource LOCBattleNetLinksPCGW}", @"http://pcgamingwiki.com/w/index.php?search=" + product.Name));
```
JosefNemec: Please do actually test changes you make to see if they work. This will definitely not work and it's not the way you load dynamic resources from the code behind. You need to use `Resources.GetString` method from `IPlayniteAPI` object that's injected into a plugin.

source/Playnite/Localization/LocSource.xaml
darklinkpower: reverted to static string

source/Plugins/BattleNetLibrary/BattleNetMetadataProvider.cs
darklinkpower: Figures, I was thinking it wouldn't be that easy. I think I'll need asistance with this since I don't know exactly how it should be modified. 

For example, from my understanding it should go from this:

`           gameInfo.Links.Add(new Link("{DynamicResource LOCExampleString}", @"http://pcgamingwiki.com/w/index.php?search=" + product.Name));`

To this:
`           gameInfo.Links.Add(new Link((PlayniteApi.Resources.GetString("LOCExampleString")), @"http://pcgamingwiki.com/w/index.php?search=" + product.Name));`

I'll see if I can figure out how to set git and how to build playnite to actually test my PRs and check that they actually work.

JosefNemec: Yes `PlayniteApi.Resources.GetString` is the right use, but you need to inject PlayniteApi intoBattleNetMetadataProvider first via dependency injection.

darklinkpower: If it's not too much, can you give me an example how to inject the PlayniteApi? I thought that by having `using Playnite.SDK;` but it looks like I was wrong.

source/Playnite/Localization/LocSource.xaml
```diff
@@ -777,6 +781,11 @@ Note that the Account Name option only works for accounts whose game libraries h
     <sys:String x:Key="LOCSteamCustomUrlName">Account URL Name</sys:String>
     <sys:String x:Key="LOCSteamLoggedInUser">Authenticated User</sys:String>
     <sys:String x:Key="LOCSteamRequiresApiKey">Account is private, API Key is required.</sys:String>
+    <sys:String x:Key="LOCSteamLinksCommunityHub">Community Hub</sys:String>
+    <sys:String x:Key="LOCSteamLinksDiscussions">Discussions</sys:String>
+    <sys:String x:Key="LOCSteamLinksNews">News</sys:String>
```
JosefNemec: I think News and Achievements can be general strings as well, those terms do apply to other stores as well.

darklinkpower: Also if it's possible and you have time, it'd be great to look at #1479 too, since I think they are related since it would help to have Playnite entirely in one language (at least when using steam games). And thanks for the help in this issue

PR #1487, opened 2019-10-12
Allow MSBuild from VS2019

tylerszabo:
- Prefer VS2017 when present
  - Fall back to VS2019
  - Improved error handling when no suitable MSBuild is present

### DIFF
--- a/build/common.ps1
+++ b/build/common.ps1
@@ -83,15 +83,19 @@ function global:Get-MsBuildPath()
         }
     }
 
-    $path = & $VSWHERE_CMD -version "[15.0,16.0)" -requires Microsoft.Component.MSBuild -find "MSBuild\**\Bin\MSBuild.exe"
-    if (Test-Path $path)
+    $path = & $VSWHERE_CMD -version "[15.0,16.0)" -requires Microsoft.Component.MSBuild -find "MSBuild\**\Bin\MSBuild.exe" -latest | Select-Object -First 1
+    if ($path -and (Test-Path $path))
     {
         return $path
     }
-    else
+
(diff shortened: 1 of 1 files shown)


build/common.ps1
```diff
@@ -83,15 +83,19 @@ function global:Get-MsBuildPath()
         }
     }
 
-    $path = & $VSWHERE_CMD -version "[15.0,16.0)" -requires Microsoft.Component.MSBuild -find "MSBuild\**\Bin\MSBuild.exe"
-    if (Test-Path $path)
+    $path = & $VSWHERE_CMD -version "[15.0,16.0)" -requires Microsoft.Component.MSBuild -find "MSBuild\**\Bin\MSBuild.exe" -latest | Select-Object -First 1
```
tylerszabo: Prefer VS2017 to avoid surprises for anyone running both version

build/common.ps1
```diff
@@ -83,15 +83,19 @@ function global:Get-MsBuildPath()
         }
     }
 
-    $path = & $VSWHERE_CMD -version "[15.0,16.0)" -requires Microsoft.Component.MSBuild -find "MSBuild\**\Bin\MSBuild.exe"
-    if (Test-Path $path)
+    $path = & $VSWHERE_CMD -version "[15.0,16.0)" -requires Microsoft.Component.MSBuild -find "MSBuild\**\Bin\MSBuild.exe" -latest | Select-Object -First 1
+    if ($path -and (Test-Path $path))
```
tylerszabo: Check `$path` is not null, first, sincel `Test-Path` will slightly obscure error "Cannot bind argument to parameter 'Path' because it is null."

PR #1489, opened 2019-10-13
Add FinalBurn Neo profile to RetroArch emulator definition

bsinky:
Originally had the profile name as "FB Neo" to better match the related cores, but then I realized RetroArch's Online Updater calls it "FinalBurn Neo" so I matched that instead.

Didn't place it in the list alphabetically since I wanted to keep it by the other FB/FinalBurn cores.


JosefNemec:
RetroArch profiles should be updated via [GenerateRetroArchProfile.ps1](https://github.com/JosefNemec/Playnite/blob/master/scripts/GenerateRetroArchProfile.ps1) script, not manually. It will automatically generate profiles for all supported cores. I'll do it for Playnite 6 release.

Issue #1481, opened 2019-10-02
Hours Played Sync


absimilliard:
Hi, the hours played still do not sync properly with Steam, and does not sync at all with the other platforms.

JosefNemec:
Only GOG, Steam and Origin currently support play time sync: https://github.com/JosefNemec/Playnite/wiki/Feature-table-based-on-library-provider (with [GOG being bugged](https://github.com/JosefNemec/Playnite/issues/1474) at the moment). You also have to have import of uninstalled games enabled.

Also once Playnite recorded/imported play time it will no longer keep syncing it and use internal time tracking instead. Unless you enable "Force sync play time" in the settings.

Issue #1474, opened 2019-09-30
"Force sync play time" doesn't seem to apply to the GOG library anymore


ClearStrelok:
I thought this could be an issue on my end, but I've had no issue syncing with Steam.

JosefNemec:
We don't distinguish between libraries when deciding whether to download time or not. It must mean that we are not getting play time from GOG at all. Is play time downloaded for a game if you set it to 0?

ClearStrelok:
It doesn't, but I will reinstall Galaxy and try again to make sure I'm not causing this.

JosefNemec:
How do you link the GOG account? Via account name or by authenticating? Account name sync has been quite buggy (on the GOG's side) and I will be removing it in Playnite 6.

ClearStrelok:
I'm using the account authentication. Right now, syncing is still broken after reinstalling GOG Galaxy and purging the web cache so I'm guessing this is an API issue. Could this be related to the ongoing beta of Galaxy 2.0 ?

JosefNemec:
Galaxy installation will have no impact on this. We get the info from GOG's web API.

JosefNemec:
I'm afraid I have to close this without a fix because I had to switch to a different API to get list of games (it's more reliable) and sadly that API doesn't provide play time information.

PR #1494, opened 2019-10-19
Modify Loc Strings related to playtime sync

darklinkpower:
I propose modifying three strings to be more clear:

1. `Force sync play time` -> `Force sync Play Time for supported libraries` (Makes it clear that not every library is supported)
2. `Always download recorded play time from remote library during library sync, ignoring time recorded by Playnite.` -> 
`Always download recorded Play Time from remote library during library sync, ignoring time recorded by Playnite. GOG, Origin and Steam libraries are supported when the user is authenticated.` (To specify exactly what libraries are supported and to explain that the user must be currently authenticated in each respective library for Playtime Sync to work.
3. `Playtime is tracked and stored in seconds.` -> `Play Time is tracked and stored in seconds.` (For consistency since the other strings have a space between `Play` and `Time`. I also capitalized the first letter of each word.

Just as another suggeston, I was thinking that the terms `Play Time` and `Time Played` in the different strings could be unified using one of the two, since they have the same meaning. If this was the case, I'm more inclined to use `Time Played`

I propose this because I've seen confusion on which libraries can sync playtime and some playtime sync problems that have been reported were because the user was not authenticated, as seen in #1481 and on discord different times. It should also be mentioned that the GOG sync is currently not working as mentioned in #1474

### DIFF
--- a/source/Playnite/Localization/LocSource.xaml
+++ b/source/Playnite/Localization/LocSource.xaml
@@ -372,8 +372,8 @@ New database location will only be used after application is restarted.</sys:Str
     <sys:String x:Key="LOCSteamBackgroundSourceScreenshot">Store Screenshot</sys:String>
     <sys:String x:Key="LOCSteamBackgroundSourceStore">Store Background</sys:String>
     <sys:String x:Key="LOCBackgroundImageScreenOptionTooltip" xml:space="preserve">Doesn't apply retrospectively to existing games without re-downloading metadata.</sys:String>
-    <sys:String x:Key="LOCSettingsForceDownloadPlaynite">Force sync play time</sys:String>
-    <sys:String x:Key="LOCSettingsForceDownloadPlayniteTooltip">Always download recorded play time from remote library during library sync, ignoring time recorded by Playnite.</sys:String>
+    <sys:String x:Key="LOCSettingsForceDownloadPlaynite">Force sync Play Time for supported libraries</sys:String>
+    <sys:String x:Key="LOCSettingsForceDownloadPlayniteTooltip">Always download recorded Play Time from remote library during library sync, ignoring time recorded by Playnite. GOG, Origin and Steam libraries are supported when the user is authenticated.</sys:String>
     <sys:String x:Key="LOCSettingsXInputInDesktopMode">Enable controller support in Desktop mode</sys:String>
     <sys:String x:Key="LOCSettingsXInputGuideOpensFullscreen">Guide button opens Fullscreen mode</sys:String>
     <sys:String x:Key="LOCSettingsDefaultMetadataDescription">Configure settings used when metadata are being downloaded automatically after new games are imported.</sys:String>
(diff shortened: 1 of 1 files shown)


source/Playnite/Localization/LocSource.xaml
```diff
@@ -372,8 +372,8 @@ New database location will only be used after application is restarted.</sys:Str
     <sys:String x:Key="LOCSteamBackgroundSourceScreenshot">Store Screenshot</sys:String>
     <sys:String x:Key="LOCSteamBackgroundSourceStore">Store Background</sys:String>
     <sys:String x:Key="LOCBackgroundImageScreenOptionTooltip" xml:space="preserve">Doesn't apply retrospectively to existing games without re-downloading metadata.</sys:String>
-    <sys:String x:Key="LOCSettingsForceDownloadPlaynite">Force sync play time</sys:String>
-    <sys:String x:Key="LOCSettingsForceDownloadPlayniteTooltip">Always download recorded play time from remote library during library sync, ignoring time recorded by Playnite.</sys:String>
+    <sys:String x:Key="LOCSettingsForceDownloadPlaynite">Force sync Play Time for supported libraries</sys:String>
+    <sys:String x:Key="LOCSettingsForceDownloadPlayniteTooltip">Always download recorded Play Time from remote library during library sync, ignoring time recorded by Playnite. GOG, Origin and Steam libraries are supported when the user is authenticated.</sys:String>
```
JosefNemec: I don't think that play time should be capitalized. Also I don't like listing specific libraries here, because if we add support for other in future, I'm pretty sure we will forget to update this.

source/Playnite/Localization/LocSource.xaml
```diff
@@ -372,8 +372,8 @@ New database location will only be used after application is restarted.</sys:Str
     <sys:String x:Key="LOCSteamBackgroundSourceScreenshot">Store Screenshot</sys:String>
     <sys:String x:Key="LOCSteamBackgroundSourceStore">Store Background</sys:String>
     <sys:String x:Key="LOCBackgroundImageScreenOptionTooltip" xml:space="preserve">Doesn't apply retrospectively to existing games without re-downloading metadata.</sys:String>
-    <sys:String x:Key="LOCSettingsForceDownloadPlaynite">Force sync play time</sys:String>
-    <sys:String x:Key="LOCSettingsForceDownloadPlayniteTooltip">Always download recorded play time from remote library during library sync, ignoring time recorded by Playnite.</sys:String>
+    <sys:String x:Key="LOCSettingsForceDownloadPlaynite">Force sync Play Time for supported libraries</sys:String>
```
JosefNemec: I like the addition of "supported library". Could you also change it from "Force sync" to "Prioritize"? I think that would be a better description.

source/Playnite/Localization/LocSource.xaml
darklinkpower: I'd make sure to update the string and send a PR as soon as a new library gets play time support. I think there should be a mention somewhere within Playnite of what libraries support this function but I'll remove it then but tell me if you are ok with it to add it back.

source/Playnite/Localization/LocSource.xaml
darklinkpower: So from `Force sync Play Time for supported libraries` to `Prioritize play time of supported libraries`? I'm not too sure myself if that string is self-explanatory enough to say what it does, in my opinion the word "sync" shouldn't be removed. I have a few suggestions, tell me what you think or if you still prefer your suggestion:

1. Synchronize play time of supported libraries
2. Synchronize play time of supported libraries on game library update
3. Synchronize play time of supported libraries when updating game libraries

I see there's a mention of "library during library sync" in the tooltip, but in my opinion that information goes better in the option name itself.

I'll be modifying the string to your suggestion for the moment but tell me if you like some of my suggestions.

darklinkpower: Got it, everything is modified exactly as you suggested.

JosefNemec:
Can you please switch this PR to target `devel` branch? Thanks

Issue #448, opened 2018-01-25
Support for RTL Languages 


boudrar:
Is it possible to add support for Right-To-Left languages (Persian, Hebrew and Arabic).

Using RTL localizations without changing the layout of the UI is as very bad UX.

JosefNemec:
Adding high complexity tag. We can currently simply switch application flow to RTL, but not all controls work with it properly, for example game description is upside down:

![image](https://user-images.githubusercontent.com/3874087/35432615-e66b5fc8-0280-11e8-930f-142fb975fc76.png)

Continous:
> game description is upside.

What? Why and how did that happen. Maybe if issue #397 is fixed, the game description will be fixed.

JosefNemec:
Closing, was done via #1523

PR #1523, opened 2019-11-12
Add Hebrew and RTL Support

elisherer:
I have verified that:

Adds Hebrew language and Fixes Issue #448
(I solved the issue of left-side-right and up-side-down by setting some components to have always a left-to-right direction flow)

image
image
image
Fullscreen:
image


source/Playnite.FullscreenApp/Controls/ListBoxEx.cs
```diff
@@ -21,6 +21,9 @@ public ListBoxEx() : base()
         {
             SelectionChanged += ListBoxEx_SelectionChanged;
             this.GotFocus += ListBoxEx_GotFocus;
+
+            // Set to always left-to-right
```
JosefNemec: Why is this being set always left-to-right?

source/Playnite/Localization.cs
```diff
@@ -46,6 +46,16 @@ public static string CurrentLanguage
             get; private set;
         } = SourceLanguageId;
 
+        public static CultureInfo CultureInfo
```
JosefNemec: I don't see CultureInfo and IsRightToLeft used anywhere. What are these for?

source/Playnite/Playnite.csproj
```diff
@@ -445,6 +445,11 @@
       <SubType>Designer</SubType>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
+    <None Include="Localization\he_IL.xaml">
```
JosefNemec: Please remove new language file from this PR (change in csproj and the xaml itself). It has to be merged from PR automatically generated by Crowdin, otherwise these changes will be overwritten next time Crowdin sync is made.

source/Playnite/Localization.cs
```diff
@@ -127,6 +137,10 @@ public static void SetLanguage(string language)
                 }
 
                 dictionaries.Add(res);
+
+                CultureInfo culture = new CultureInfo(language.Replace("_", "-"), false);
+                CultureInfo.DefaultThreadCurrentCulture = culture;
```
JosefNemec: I'm not sure if it's a good idea to force application culture based on the app language. For example I personally use Playnite in English, but I also use Czech system culture and I want to have dates and times in Czech in Playnite. I think we should be preferring system culture. But please correct me if this is here for some other reason that I'm not aware of for Hebrew support specifically.

Also this will not work for default English language and system culture will be used (the if statement will not be called in that case).

source/Playnite.FullscreenApp/Controls/ListBoxEx.cs
elisherer: Actually I thought it made a difference in the up-side-down symptom, but I removed it and it didn't. So reverted.

source/Playnite/Localization.cs
elisherer: At first I thought of storing the CultureInfo only in Localization instead of the thread. But if you think it would be better not to set the thread culture, then it's back to these properties to be queried instead.

source/Playnite/Localization.cs
elisherer: I personally use English system culture and hebrew dates are formatted differently from English.
But I don't mind have it how you described it. I will create a local CultueInfo object that will be used to query for culture properties (i.e RTL)

Issue #1550, opened 2019-11-24
Change term "Scripts" to "Global Scripts" in the main settings window.


darklinkpower:
**Describe the solution you'd like**
`Global Scripts` makes more sense to be more clear that the scripts written in here will apply to every game.

**Screenshots**
Proposed change:
![imagen](https://user-images.githubusercontent.com/1389286/69488178-78208080-0e2b-11ea-81ac-1430bef409f5.png)

JosefNemec:
I would rather add some description to the settings view then change this string, I'll do it for 6.3.

PR #1551, opened 2019-11-24
Change term "Scripts" to "Global Scripts" in the main settings window. enhancement 

darklinkpower:
I have verified that:

For #1550


JosefNemec:
I would rather add some description to the settings view then change this string, I'll do it for 6.3.

PR #1684, opened 2020-03-01
Separate Bizhawk and Mednafen emulator profile by platforms in emulator definitions file and standarize some platform names

darklinkpower:
I have verified that:

And small fix to the bsnes-hd profile.

### DIFF
--- a/source/Playnite/Emulators/Definitions.yaml
+++ b/source/Playnite/Emulators/Definitions.yaml
@@ -549,7 +549,7 @@
   Profiles:
     - Name: Default
       DefaultArguments: '"{ImagePath}"'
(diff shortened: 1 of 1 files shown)


JosefNemec:
Did you check that platform names are compatible what Retroarch uses? They sometimes use different format and that could cause duplicates.

The easiest way to check is to have Playnite started with clean library and check Platforms list in library manager since we build that list from emulator profiles.

darklinkpower:
Thanks for the heads up. I copied the platform names as they were and just separated them by platforms so there shouldn't be an issue. But that being said... I just checked and there's an issue with wonderswan and wonderswan color having two different names in different emulators so I copied something that was wrong.

`Bandai Wonderswan` (By Retroarch) and `Wonderswan` (By Mednafen and Bizhawk)
`Bandai WonderSwan Color` (By Retroarch) and `Bandai WonderSwan Color` (By Mednafen and Bizhawk)

Perhaps there are more mistakes like this. I'll be checking tomorrow and fix what I find and take retroarch as standard for naming.

darklinkpower:
I separated the Mednafen profile by platforms too and I didn't find any other issue so it should be good to go.

PR #1778, opened 2020-04-29
Add new include_free_sub parameter for Steam GetOwnedGames API call.

psychonic:
Earlier this week, Steam added a new parameter to their IPlayerService/GetOwnedGames Web API call to opt into including free games that everyone owns. By adding this, more uninstalled games can be synced.

In testing, it does pull at least some more for me.

See: https://github.com/SteamDatabase/SteamTracking/commit/e84ef2b4be1a9f32f50dc5b7e07fd6553dbe11fa#diff-1dd8848646dd2385514346656345b4d4R90 or https://api.steampowered.com/ISteamWebAPIUtil/GetSupportedAPIList/v1?key=KEYHERE

I have verified that:

### DIFF
--- a/source/PlayniteServices/Controllers/Steam/LibraryController.cs
+++ b/source/PlayniteServices/Controllers/Steam/LibraryController.cs
@@ -83,7 +83,7 @@ namespace PlayniteServices.Controllers.Steam
             var steamId = ulong.TryParse(userName, out var directId) ? directId.ToString() : GetUserId(userName);
 
             var libraryUrl = string.Format(
-                @"http://api.steampowered.com/IPlayerService/GetOwnedGames/v0001/?key={0}&include_appinfo=1&format=json&steamid={1}&include_played_free_games=1",
+                @"http://api.steampowered.com/IPlayerService/GetOwnedGames/v0001/?key={0}&include_appinfo=1&format=json&steamid={1}&include_played_free_games=1&include_free_sub=1",
                 Steam.ApiKey, steamId);
             WaitRequest();
 
--- a/source/Plugins/SteamLibrary/SteamLibrary.cs
+++ b/source/Plugins/SteamLibrary/SteamLibrary.cs
@@ -369,7 +369,7 @@ namespace SteamLibrary
 
         internal GetOwnedGamesResult GetPrivateOwnedGames(ulong userId, string apiKey)
         {
-            var libraryUrl = @"https://api.steampowered.com/IPlayerService/GetOwnedGames/v0001/?key={0}&include_appinfo=1&include_played_free_games=1&format=json&steamid={1}";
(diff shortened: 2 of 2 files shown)


JosefNemec:
I remember that we used this in the past, but I specifically removed it to fix some issue. I think it was importing some free games even if user never played or installed them, but I don't remember exactly.

psychonic:
I believe this specific parameter is new, but maybe there was previously some similar functionality.

That seems like a valid concern. Some users would want all games they own (even as free), regardless of whether they've installed or played it before. Some would prefer what Steam does now, not displaying free games you haven't played.

JosefNemec:
What kind of games does this affect specifically (do you have some examples)? All free games on Steam? How is this it different to `include_played_free_games`?

psychonic:
I believe the results will be different for each person. For me, there are only 18 more apps returned with it than without. Many of them are older Source engine mods.

It does not include all free games on Steam, most of which have their own packages that are just marked as free-on-demand. Instead, it includes whatever is in [package 0](https://steamdb.info/sub/0/apps/). So what gets added is any game listed there that the user has not played, or has not played since Steam started tracking play time per OS. I have an example of that last case, where the API even tells me I have playtime in the GetOwnedGames response, but the windows/mac/linux playtime values are 0 and the app is not listed without include_free_sub.

```
appid                       : 17530
name                        : D.I.P.R.I.P. Warm Up
playtime_forever            : 5
img_icon_url                : 3f8cb122a58b0a0a35bbdabb0d17a34e057909fc
img_logo_url                : 4ae37a0ba8f01c14e9a5a8a929f0fdfb10f60f22
has_community_visible_stats : True
playtime_windows_forever    : 0
playtime_mac_forever        : 0
playtime_linux_forever      : 0
```

JosefNemec:
Do you think this could import a lot of "garbage" that people would not want imported? Would it be beneficial adding this as an option then?

psychonic:
I can't speak for others, but I do appear to be in the minority that would prefer to see more than less, just managing visibility with favorites / categories / filters. And I'm just using a fork of the extension now anyway for this and a few other things.

I don't think it's worth adding as an option due to how little it actually adds.

JosefNemec:
I'm not against including this. I think we should go by simple rule: Are these games visible in Steam client library? If so, then we should include them.

Also, I haven't checked, but we may start importing these once #1808 is done anyways.

psychonic:
Oddly, the games included in the package do not display in the library unless installed (but all show as owned in the store). The tools included in the package do show in the library even when not installed.

JosefNemec:
Can you please make a new PR into `devel8` branch? Playnite 8 will have an option to ignore games on import, so this shouldn't cause any issue if we start importing unwanted stuff, people can just ignore it.

PR #1799, opened 2020-05-07
Update logic for checking if Humble games have keys for another service

### LINKED ISSUE #1797: Importing some third party store games from Humble when set to ignore them
**Check for existing issue**
This was sparked by #1676 and is probably something to improve while you're working on it.

**Describe the bug**
I really appreciate Playnite and took advantage of the Humble integration to run the importer and bring more of my games together. To my surprise, even though I had the "Ignore third party store games" checkbox checked, the importer imported _Assault Android Cactus+_, _FRAMED Collection_, and _Into the Breach_, which I also have Steam keys for. To reproduce, own any of those games on Humble and import from Humble into Playnite with the "Ignore third party store games" setting enabled.

**Fix**
After mucking around in Humble's API response and the source code for a bit, it seems that each of these games has a separate problem. From reading it, it looks like HumbleLibrary.cs just checks if the store key's name is the same as the game's or has " Key" appended to the end. However, "Assault Android Cactus+" != "Assault Android Cactus", "FRAMED Collection" != "Framed Collection", and "Into the Breach" != "Into The Breach (Steam)", which are the differences between the titles of the downloadable application and the Steam key. We could fix the last two by comparing the strings case-insensitively and also checking if a "Steam" is appended to the end, but I'm not sure how to fix the first. I'd be willing to work on this before the 8.0 release, though.

**Diagnostics ID**
9666630c-382d-47a4-bc0e-808c4618efb8, if you need it.

awbooze:
I have verified that:

Closes #1797 when completed, but not ready to merge yet, which is why it's a draft.

### DIFF
--- a/source/Plugins/HumbleLibrary/HumbleLibrary.cs
+++ b/source/Plugins/HumbleLibrary/HumbleLibrary.cs
@@ -132,7 +132,10 @@ namespace HumbleLibrary
                                 {
                                     var exst = allTpks.FirstOrDefault(a =>
                                     !a.human_name.IsNullOrEmpty() &&
-                                    (a.human_name == product.human_name || Regex.IsMatch(a.human_name, product.human_name + @".+\sKey$")));
+                                    (a.human_name.Equals(product.human_name, StringComparison.OrdinalIgnoreCase) || 
+                                    Regex.IsMatch(a.human_name, product.human_name + @".+\sKey$", RegexOptions.IgnoreCase) || 
+                                    Regex.IsMatch(a.human_name, product.human_name + @".*\s\(?Steam\)?$", RegexOptions.IgnoreCase)));
+
                                     if (exst != null)
                                     {
                                         continue;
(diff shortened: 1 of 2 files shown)


awbooze:
Alright, I've fixed all the issues on my end. I'd appreciate some testing though so that we can be sure that these changes don't remove anything from other people's libraries that isn't available on a third-party platform.

source/Plugins/HumbleLibrary/HumbleLibrary.cs
```diff
@@ -132,7 +132,11 @@ public override IEnumerable<Game> ImportGames()
                                 {
                                     var exst = allTpks.FirstOrDefault(a =>
                                     !a.human_name.IsNullOrEmpty() &&
-                                    (a.human_name == product.human_name || Regex.IsMatch(a.human_name, product.human_name + @".+\sKey$")));
+                                    (a.human_name.Equals(product.human_name, StringComparison.OrdinalIgnoreCase) || 
+                                    Regex.IsMatch(a.human_name, product.human_name + @".+\sKey$", RegexOptions.IgnoreCase) || 
+                                    Regex.IsMatch(a.human_name, product.human_name + @".*\s\(?Steam\)?$", RegexOptions.IgnoreCase) ||
+                                    Regex.IsMatch(a.human_name + @"\s*+$", product.human_name, RegexOptions.IgnoreCase)));
```
JosefNemec, line 138: What is this last check for?

awbooze: That's checking for the problem I had with Assault Android Cactus. For that game, a.human_name == "Assault Android Cactus", while product.human_name == "Assault Android Cactus+". The regex checks if there are any number of spaces and then a "+" symbol, just in case.

Issue #1796, opened 2020-05-06
Easy way to change the completion status


Skypper67:
It would be very interesting if we could change the completion status with less clicks.

For example, having an option inside the context menu to change, instead of edit a game and then go to Advanced.

JosefNemec:
Done in Playnite 8.

PR #1814, opened 2020-05-15
add submenu for changing completion status to game context menu (#1796)

V0idhead:
Since this is my first contribution I wanted to make sure I can get everything to work before making promises on the ticket. So sorry for the "surprise" pull request.

As proposed in https://github.com/JosefNemec/Playnite/issues/1796 I added an item in the game context menu to change the completion status:
completionStatus

If this change is welcome I'd also make a similar change to the version of the context menu that handles multiple selected games.

Obviously the Header should be "Set Completion Status", but I don't know the process for updating the localizations so I reused the existing string.

I have verified that:

* [x]  These changes work, by building the application and testing them.
* [x]  Pull request is targeting `devel` branch.
* [ ]  I added myself into [contributors file](https://github.com/JosefNemec/Playnite/blob/devel/source/Playnite.DesktopApp/Resources/contributors.txt) if I want to receive contribution credit in the application itself.

### DIFF
--- a/source/Playnite/GamesEditor.cs
+++ b/source/Playnite/GamesEditor.cs
@@ -328,6 +328,12 @@ namespace Playnite
             }
         }
 
+        public void SetCompletionStatus(Game game, CompletionStatus status)
+        {
+            game.CompletionStatus = status;
+            Database.Games.Update(game);
+        }
+
         public void RemoveGame(Game game)
         {
             if (game.IsInstalling || game.IsRunning || game.IsLaunching || game.IsUninstalling)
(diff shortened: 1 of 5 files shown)


JosefNemec:
To add a new translatable string, create new entry in `LOCSource.xaml` file.

Also, please add check list from the PR template.

source/Playnite/GamesEditor.cs
```diff
@@ -328,6 +328,12 @@ public void ToggleFavoriteGame(List<Game> games)
             }
         }
 
+        public void SetCompletionStatus(Game game, CompletionStatus status)
+        {
+            game.CompletionStatus = status;
```
JosefNemec, line 333: Can you add check if new status is actually different and skip game update if it's the same?

V0idhead:
Sure,

 - I added an entry LOCSource.xaml and also added an entry to the german version
 - added the checklist
 - made sure the new status is different before updating the game
 - also added the menu to the version of the context-menu that handles multiple games

PR #1833, opened 2020-05-21
Covers aspect ratio fixes.

darklinkpower:
I have verified that:

To fix a few things:
1. GOG recently changed the recommended resolution and aspect ratio to 660x930, which has an aspect ratio of 22:31
2. Fix Epic cover aspect ratio that actually uses 3:4
3. Change the `Steam New UI` string to `Steam vertical cover` to use the same term used in other places like the Steam library settings and Steam Store metadata plugin settings.


JosefNemec:
- Did you check that `LOCSettingsCovertAspectSteamNewUI` is not used anywhere else since you are removing it?
- Do you have link where GOG mentions the change?
- I don't think there's a uniform Epic Store aspect ratio. Just checking their store and they use multiple aspect ratios and sizes: 770×854, 640 × 854, 759 × 854. It's a mess. However, it's true that they scale it into 3:4.

darklinkpower:
1. Positive. It's only used in `source/Playnite.DesktopApp/Controls/SettingsSections/AppearanceGridView.xaml` and the localization files.

2. It shows as of the more recent client update. I think they made the change because the previous aspect ratio had some strange numbers.
imagen


3. I went with how it looks in the Epic client library cover view and in the store
imagen
imagen

Issue #136, opened 2017-10-08
Choosing the language for the game details


XploitOcelot:
It would be great to be able to choose in wich language I want to have the downloaded details. I'm a spaniard, but live in Switzerland.
My option would be to have the details in either spanish or english. BUT for Steam and Origin games the details are downloaded only in english, from GOG everything is in german... it's quite a mess.

It's a suggestion that sooner or later would come.
I'm totally in love with Playnite and only use it from yesterday. THANK YOU!!!

darklinkpower:
Didn't find this issue when I made my post, my bad. I'll copy paste the info regarding adding steam metadata support in other languages.

> **Describe the solution you'd like**
> Currently, Steam grabs metadata in english language only even if you use Playnite in any other language. 
> 
> From what I see, Playnite currently pulls metadata via the Store API without any language in specific in `SteamLibrary/Services/WebApiClient.cs` this way:
> 
> `https://store.steampowered.com/api/appdetails?appids={appId}`
> 
> **Adding the `l` parameter, outputs the description in the target language:**
> `https://store.steampowered.com/api/appdetails?appids={appId}&l={SteamApiLangCode}`
> 
> **For example:**
> [https://store.steampowered.com/api/appdetails?appids=460790&l=japanese](https://store.steampowered.com/api/appdetails?appids=460790&l=japanese)
> 
> Perhaps the code could be changed to `https://store.steampowered.com/api/appdetails?appids={appId}&l=` and just add the target language via a localization string per language.
> 
> The Codes used by the API can be looked here (third column): 
> [https://partner.steamgames.com/doc/store/localization](https://partner.steamgames.com/doc/store/localization)
> 
> The data in the description language is still in english when the game doesn't have a description in the requested language.

And linking another issue that may or may not be related to this: #1480

rgponce:
I would like to extend this request.
It would be interesting if we could add translations to some file in the Playnite installation folder for common terms in some fields in the Details panel. Like the Genres and tags fields.
An XML file would be ideal

JosefNemec:
My current stance on this is that I don't think it's possible to implement this, for one main reason:  most of the stores and IGDB don't have localized data. So you would end up with weird mashup of localized and non-localized fields, which would also cause localized and non-localized duplicates.

If we are currently downloading some non-English fields, then we should fix that for consistency, but there's just no way we can provide consistent localized content for things that are coming from stores or game databases.

shakeyourbunny:
[...]

PR #1848, opened 2020-05-29
Choosing the language for the game descriptions (#136)

Lacro59:
This modification is a way to propose the language selection for the store with a localization (#136).
Surely not the best.

Only for Steam, Origin and Epic store.

GOG script is already localized (perhaps with default).
It's possible to force a selectable localization.
I do it in my "SuccessStory plugin".

The others scripts uses IGDB.

### DIFF
--- a/source/Plugins/EpicLibrary/EpicLibrary.cs
+++ b/source/Plugins/EpicLibrary/EpicLibrary.cs
@@ -23,11 +23,39 @@ namespace EpicLibrary
         internal readonly string TokensPath;
         internal readonly EpicLibrarySettings LibrarySettings;
 
+        // For acces to staic function from EpicAccountClient / EpicMetadataProvider // WebStoreClient
+        public static string _EpicLang { get; set; }
+
         public EpicLibrary(IPlayniteAPI api) : base(api)
         {
             playniteApi = api;
-            LibrarySettings = new EpicLibrarySettings(this, api);
+            LibrarySettings = new EpicLibrarySettings(this, api)
+            {
+                EpicLangs = GetEpicLangs()
+            };
             TokensPath = Path.Combine(GetPluginUserDataPath(), "tokens.json");
+            _EpicLang = LibrarySettings.EpicLang;
+        }
+
+        // Initialize available language list for Steam if exist in Playnite language.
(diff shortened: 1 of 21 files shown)


JosefNemec:
Thank for PR, but:

1) I still maintain [my position](https://github.com/JosefNemec/Playnite/issues/136#issuecomment-633020489) from issue #136. This is going to create a mess in user's library with ton of localized vs. non-localized duplicates. I think we should mainly aim for consistency.

2) I don't like this approach. There should be a global application wide option for preferred metadata language (or even better just use application language) and that option should be used by metadata plugins (I think that using application language makes more sense). I don't like idea of having separate option for each and every plugin.

My main problem is really the consistency question. I don't think users will be happy when we start creating duplicates in their library.

Lacro59:
> I don't like this approach. There should be a global application wide option for preferred metadata language (or even better just use application language) and that option should be used by metadata plugins (I think that using application language makes more sense). I don't like idea of having separate option for each and every plugin.

After more reflection, add a option to localize or no is the best choice.

> I still maintain my position from issue #136. This is going to create a mess in user's library with ton of localized vs. non-localized duplicates. I think we should mainly aim for consistency.

No problem.  It will stay for my personal use.

> I don't think users will be happy when we start creating duplicates in their library.

"creating duplicates in their library", with a metadata localized? how? I'm curious.

JosefNemec:
> "creating duplicates in their library", with a metadata localized? how? I'm curious.

Not all games have metadata pulled from official stores (emulated games, Uplay, Twitch games and others) and even some de-listed Steam games will have metadata pulled from alternative sources. The main alternative source is IGDB and IGDB has metadata available only in English. So for example if you import localized "Action" genre and then another game imports English "Action" genre, you now have two "Action" genres in library.

Lacro59:
OK. I understand.
Thanks for the response.

Lacro59:
The best solution creates a MetadataPlugin that modify only game description, isn't it?

JosefNemec:
Yes, I think the best way would be creating custom plugin that would fetch description in different languages.

JosefNemec:
I'm going to close based on the previous conversation. Custom metadata plugin is the way to go.

Issue #1065, opened 2019-03-06
Support for removable devices with dynamic drive letter


Kvaell:
Hello,

Some of emulator's rom aren't synchronized anymore with playnite, playnite still list them, the roms are still in the same folder, but when launched through playnite, the rom cannot be launched, each emulator saying an error occurred while reading/loading it.

If launched from the emulator itself, no problem occurs but it doesn't work frol Playnite. 

When synchronize the rom with the emulator on playnite again, the roms appears twice as if it wasn't already synchronized. 
![error](https://user-images.githubusercontent.com/27021339/53898836-2c2d7180-4039-11e9-9152-009c0a02abf0.png)

Those roms are stored on a SD card. 

It happens mainly to emulator that I don't launch often.

JosefNemec:
This apparently happens when games are stored on removable device (like SD Cards).

JosefNemec:
Is the SD card being mounted always with the same drive letter? My guess is that this happens because the game was originally imported when SD card was mounted a `G:` and later it was mounted as something else. Is that the case?

Kvaell:
I can confirm the letter doesn't change since Redream emulator doesn't have the issue, maybe for this emulator playnite make the emulator launch the roms straight away.

Kvaell:
You are right, that's exactly what happening, but I don't know how Redream emulator is still working properly, as if the letter didn't changed.

JosefNemec:
They must have built in mechanism to check other drives if the originally passed path doesn't exists. We could in theory do the same check before passing the value to an emulator and replace the path if necessary.

Kvaell:
That would be great! Normally on the worst case only the drive changes, the folder stay the same so it should be quite fast for playnite to find on which drive it is.

plug313:
I've been having lots of problems with this. I have tons of games installed on an external hard drive and every time the drive number changes I have to manually alter their installation directory/path all over again, and one at a time which is sort of painful! would love to see a solution.

JosefNemec:
Can you check if there's a way to force Windows to mount specific devices under the same drive letter? I get your frustration, but if you tell some application to use certain folder path, and then that folder path changes, it's IMHO not really a job of that application to go and look where the folder was moved.

I'll try to add some lookup for drive letter change into post Playnite 8 update release, but I don't promise anything.

plug313:
[...]

PR #1954, opened 2020-08-10
Solving issue #1065. Playnite now checks every available drive for th…

MopsiMauser:
…e given path for Images/ROMs, emulator executables and manually added games if the respective path does not exist by using the given path and just changing the drive when a game is about to be started

I have verified that:
- [x ] These changes work, by building the application and testing them.
- [x ] Pull request is targeting `devel` branch.

### DIFF
--- a/source/Playnite.Common/FileSystem.cs
+++ b/source/Playnite.Common/FileSystem.cs
@@ -326,48 +326,45 @@ namespace Playnite.Common
             }
         }
 
-        public static bool CheckDrivesForValidFilePath(ref string filePath, bool errorIfNotFullPath)
+        public static string LookupAlternativeFilePath(string filePath, bool errorIfNotFullPath)
         {
-            return CheckDrivesForValidFilePath(new FileInfo(filePath), out filePath, errorIfNotFullPath);
+            return CheckDrivesForPath(filePath, errorIfNotFullPath, path => File.Exists(path));
         }
 
-        public static bool CheckDrivesForValidFilePath(FileInfo fileInfo, out string filePath, bool errorIfNotFullPath)
+        public static string LookupAlternativeDirectoryPath(string directoryPath, bool errorIfNotFullPath)
         {
-            filePath = fileInfo.FullName;
+            return CheckDrivesForPath(directoryPath, errorIfNotFullPath, path => Directory.Exists(path));
+        }
 
-            if (!Paths.IsFullPath(fileInfo.FullName))
+        private static string CheckDrivesForPath(string originalPath, bool errorIfNotFullPath, Predicate<string> predicate)
+        {
+            if (!Paths.IsFullPath(originalPath))
             {
                 if (errorIfNotFullPath)
                 {
-                    throw new ArgumentException("Provided path is not rooted", nameof(filePath));
+                    throw new ArgumentException("Provided path is not rooted", nameof(originalPath));
                 }
                 else
                 {
-                    return true;
+                    return originalPath;
                 }
             }
 
-            if (fileInfo.Exists)
-            {
-                return true;
-            }
-
-            var rootPath = Path.GetPathRoot(fileInfo.FullName);
+            var rootPath = Path.GetPathRoot(originalPath);
             var availableDrives = DriveInfo.GetDrives()
                 .Where(d => d.IsReady && !d.Name.Equals(rootPath, StringComparison.OrdinalIgnoreCase));
 
-            foreach(var drive in availableDrives)
+            foreach (var drive in availableDrives)
             {
-                var filePathWithoutDrive = fileInfo.FullName.Substring(drive.Name.Length);
-                var newFilePath = Path.Combine(drive.Name, filePathWithoutDrive);
-                if (File.Exists(newFilePath))
+                var pathWithoutDrive = originalPath.Substring(drive.Name.Length);
+                var newDirectoryPath = Path.Combine(drive.Name, pathWithoutDrive);
+                if (predicate(newDirectoryPath))
                 {
-                    filePath = newFilePath;
-                    return true;
+                    return newDirectoryPath;
                 }
             }
 
-            return false;
+            return string.Empty;
         }
     }
 }
--- a/source/Playnite.Common/FileSystem.cs
+++ b/source/Playnite.Common/FileSystem.cs
@@ -364,7 +364,7 @@ namespace Playnite.Common
             }
             catch (Exception ex) when (!Debugger.IsAttached)
             {
-                logger.Error(ex, "Error looking for alternative path");
+                logger.Error(ex, $"Error looking for alternative path for original path \"{originalPath}\"");
             }
 
             return string.Empty;
--- a/source/Playnite.Common/FileSystem.cs
+++ b/source/Playnite.Common/FileSystem.cs
@@ -9,6 +9,7 @@ using System.Threading.Tasks;
 using System.Security.Cryptography;
 using System.IO.Compression;
 using Playnite.SDK;
+using System.Diagnostics;
 
 namespace Playnite.Common
 {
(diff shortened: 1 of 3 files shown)


source/Playnite/Controllers/GenericGameController.cs
```diff
@@ -133,5 +145,59 @@ private void Monitor_TreeDestroyed(object sender, EventArgs args)
...
+                    logger.Warn($"Path \"{gameAction.Path}\" does not exist for game \"{Game.Name}\"" +
+                        $" and is temporarily changed to {gameActionPath}");
+                    gameAction.Path = gameActionPath;
+                }
+            }
+        }
+
+        private void CheckEmulatorConfigExecutable(EmulatorProfile emulatorProfile)
+        {
+            if (Game.PlayAction.Type == GameActionType.Emulator)
```
JosefNemec, line 187: Working directory has to be handled as well. Some Play actions have relative Path, but use absolute WorkingDir instead. The same goes for emulators.

source/Playnite/Controllers/GenericGameController.cs
```diff
@@ -133,5 +145,59 @@ private void Monitor_TreeDestroyed(object sender, EventArgs args)
             stopWatch.Stop();
             OnStopped(this, new GameControllerEventArgs(this, stopWatch.Elapsed.TotalSeconds));
         }
+
+        private void CheckGameImagePath()
```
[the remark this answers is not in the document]
MopsiMauser: So these specific check methods should just return the path which is then assigned to the Game object in the Play method? I think i need to change the Game object because it is directly used to expand the variables as far as i understand it.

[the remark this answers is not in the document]
JosefNemec: You can create copy (for example via `GetClone` extension method) and modify that copy. I just don't like the idea of modify the input object since it's probably used elsewhere up the call stack.

MopsiMauser:
Thank you for your advices. I refactored the code and added the same logic when an installation directory is about to be opened via context menu. @JosefNemec

source/Playnite.Common/FileSystem.cs
```diff
@@ -325,5 +325,46 @@ public static void CopyDirectory(string sourceDirName, string destDirName, bool
...
+        public static string LookupAlternativeDirectoryPath(string directoryPath, bool errorIfNotFullPath)
+        {
+            return CheckDrivesForPath(directoryPath, errorIfNotFullPath, path => Directory.Exists(path));
+        }
+
+        private static string CheckDrivesForPath(string originalPath, bool errorIfNotFullPath, Predicate<string> predicate)
+        {
+            if (!Paths.IsFullPath(originalPath))
+            {
+                if (errorIfNotFullPath)
```
JosefNemec, line 343: I really think we don't need to have an option to throw an error. Especially if we don't use it anywhere.

source/Playnite/Controllers/GenericGameController.cs
```diff
@@ -20,26 +21,43 @@ public class GenericGameController : BaseGameController
...
                 throw new Exception("Cannot start game without play action.");
             }
 
-            var playAction = Game.PlayAction.ExpandVariables(Game);
+            var gameClone = Game.GetClone();
+            var playAction = gameClone.PlayAction;
+            var emulators = database.Emulators.ToList();
+            var profileClone = GameActionActivator.GetGameActionEmulatorConfig(playAction, emulators)?.GetClone();
+
+            //Solving Issue #1065 if removable drives are used for storing ROMs/Images
```
JosefNemec, line 43: Can you remove references to #1065? I don't think it's necessary. This is not workaround for some issue, but new feature and I usually don't add references for those.

source/Playnite/Controllers/GenericGameController.cs
```diff
@@ -20,26 +21,43 @@ public class GenericGameController : BaseGameController
...
+        {
             if (Game.PlayAction == null)
             {
                 throw new Exception("Cannot start game without play action.");
             }
 
-            var playAction = Game.PlayAction.ExpandVariables(Game);
+            var gameClone = Game.GetClone();
+            var playAction = gameClone.PlayAction;
+            var emulators = database.Emulators.ToList();
```
JosefNemec, line 40: Why are you creating copy list of emulators? I don't see it use anywhere, but in one place, and I don't see a reason for this. Please correct me if I'm wrong.

source/Playnite/Controllers/GenericGameController.cs
```diff
@@ -20,26 +21,43 @@ public class GenericGameController : BaseGameController
...
+            {
+                //Solving Issue #1065 if removable drives are used for emulators
+                CheckEmulatorConfig(profileClone);
+            }
+
+            playAction = gameClone.PlayAction.ExpandVariables(gameClone);
 
             Dispose();
+
+            profileClone = profileClone?.ExpandVariables(gameClone);
```
JosefNemec, line 57: Can you move it above dispose call? It look weird just separated here.

source/Playnite/GamesEditor.cs
```diff
@@ -260,7 +260,19 @@ public void OpenGameLocation(Game game)
 
             try
             {
-                Process.Start(game.ExpandVariables(game.InstallDirectory));
+                var gameClone = game.GetClone();
```
JosefNemec, line 263: Correct if I'm wrong, but I don't see a reason to create game copy here.

source/Playnite.Common/FileSystem.cs
```diff
@@ -325,5 +325,46 @@ public static void CopyDirectory(string sourceDirName, string destDirName, bool
...
+                {
+                    throw new ArgumentException("Provided path is not rooted", nameof(originalPath));
+                }
+                else
+                {
+                    return originalPath;
```
JosefNemec, line 349: Should return empty string as well.

source/Playnite.Common/FileSystem.cs
```diff
@@ -325,5 +325,46 @@ public static void CopyDirectory(string sourceDirName, string destDirName, bool
...
+        {
+            return CheckDrivesForPath(filePath, errorIfNotFullPath, path => File.Exists(path));
+        }
+
+        public static string LookupAlternativeDirectoryPath(string directoryPath, bool errorIfNotFullPath)
+        {
+            return CheckDrivesForPath(directoryPath, errorIfNotFullPath, path => Directory.Exists(path));
+        }
+
+        private static string CheckDrivesForPath(string originalPath, bool errorIfNotFullPath, Predicate<string> predicate)
```
JosefNemec, line 339: Can you try/catch the whole method just to be sure? log the error, of course, but also add condition to catch only if debugger is not attached. Thanks

I don't think this should generally fail, but I've seen of lot of weird stuff happening on people's PCs, so just to be sure. Especially if this is just a bonus feature for cases where users use wrong paths in the first place.

source/Playnite/Controllers/GenericGameController.cs
MopsiMauser: I did not add ToList() call. It was already there. I guess because the GameActionActivator.GetGameActionEmulatorConfig(playAction, emulators) method needs an List<Emulator> instance as second parameter

source/Playnite/GamesEditor.cs
MopsiMauser: I am creating a copy, because I change the provided Game object. But as you mention it now, i just noticed that InstallDirectory can contain placeholder and so i need to resolve these placeholders before I can check if the path or alternatives exist

source/Playnite.Common/FileSystem.cs
MopsiMauser: Do you mean this check when (!PlayniteEnvironment.ThrowAllErrors) or actually !Debugger.IsAttached ?

MopsiMauser:
I just tested it using UNC path and I discovered a bug. When expanding the path for a GameAction Paths.FixSeparators is called which replaces \\ with \ and an invaild path is generated, "\\LAPTOP-5BNL27DP\KeePass Password Safe 2\Keepass.exe" becomes "\LAPTOP-5BNL27DP\KeePass Password Safe 2\Keepass.exe" (please don't mind that Keepass isn't actually a game. I just added it manually for testing purpose)

source/Playnite/GamesEditor.cs
JosefNemec: Just store original install directory into a variable and work with that, making copy of the game object is unnecessary.

source/Playnite/Controllers/GenericGameController.cs
JosefNemec: Hmm, ok then.

source/Playnite.Common/FileSystem.cs
JosefNemec: !Debugger.IsAttached

JosefNemec:
I was talking about mounted UNC paths, when you connect it as new drive. There are known issues when using direct UNC paths, ignore those. There's a separate issue to fix them.

source/Playnite/GamesEditor.cs
MopsiMauser: But first I need to expand the placeholders before I check for alternatives, don't I?

MopsiMauser:
Ah, ok. I didn't understand it properly. Now I checked it and it works with mounted UNC paths.

source/Playnite/GamesEditor.cs
MopsiMauser: But I think it makes actually no sense using placeholders for install direcotry because it is used to resolve placeholder {InstallDir}, which could lead to the problem that {InstallDir} is resolved with the value {InstallDir}, or am I wrong? ;-)

JosefNemec: I don't know why there's expanding going on in the first place. I think we can remove that, because it IMO doesn't make sense to have variables in installation directory (I can't think of a use case for them there). I don't remember why I added expanding here.

MopsiMauser:
@JosefNemec I refactored the code you commented and pushed it

source/Playnite.Common/FileSystem.cs
```diff
@@ -325,5 +326,48 @@ public static void CopyDirectory(string sourceDirName, string destDirName, bool
...
+                    {
+                        return newDirectoryPath;
+                    }
+                }
+
+                return string.Empty;
+            }
+            catch (Exception ex) when (!Debugger.IsAttached)
+            {
+                logger.Error(ex, "Error looking for alternative path");
```
JosefNemec, line 367: Can you log input path please? It's a last thing, I promise :)

MopsiMauser: ok, it's done ;-)

PR #2040, opened 2020-10-08
PowerShell Cleanup v2

bburky:
First pass at reimplementing #924.
* Switch to using the `Microsoft.PowerShell` API to build PowerShell commands.
* Load extensions using `Import-Module` instead of evaling the source.
* Replace PowerShell `GetFunctionExits()` with a `GetFunction()` method that returns the `CommandInfo` object.
* Enable working debugging and breakpoints
* Load PowerShell extensions as PowerShell modules

Loading the `.psm1` extension source using `Import-Module` correctly causes the correct file path to be shown while debugging an extension. Additionally, the automatic `$PSScriptRoot` and `$PSCommandPath` variables now are correctly set.

~~Because the extension is loaded as a `.ps1` file and not a `.psm1` file, it is not possible to access the `ExportedFunctions` like #924 did. Additionally, global functions never show up in `ExportedFunctions`  which are used by lots of existing extensions.~~ 

Extensions are now loaded as a PowerShell module. The exported functions are looked up using `ExportedCommands`  instead of `Get-Command`.

`PowerShellTests.ExecuteWorkDirTest` was broken, it accidentally depended on having a usable working directory from `IronPythonTests` running before it. I fixed it and the one for IronPython, but now `IronPythonTests.ExecuteWorkDirTest` is failing. If `PowerShellTests.ExecuteWorkDirTest` was run alone, the following error would occur:
```
Playnite.SDK.Exceptions.ScriptRuntimeException : Access to the path 'C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\Common7\IDE\workDirTest.txt' is denied.
```

Debugging and breakpoints somewhat to work fine now. ~~I'm not quite sure why but I can't get it to work in VS Code's PowerShell integrated debugger.~~ I _was_ able to use PowerShell ISE though. Some reason a second  "[Remote File]" tab pops up in PowerShell ISE, maybe it's not correctly finding the original source file on disk, ~~and that's why VS Code doesn't work either.~~

Debugging usage:

1. Run the following commands in the PowerShell ISE console:
```powershell
Enter-PSHostProcess -Name Playnite.DesktopApp
Debug-Runspace -Name "LibraryExporter.ps1"
```
2. Trigger your extension in Playnite somehow (e.g. click the menu to trigger `GetMainMenuItems).
3. A tab labeled "[Remote File]" will appear in PowerShell ISE. Breakpoints may be set now. Click "Run/Continue" when ready.
4. Interact with your extension some more and breakpoints will be hit.

Breaking changes:
* PowerShell extension files must be a PowerShell module. For most extensions this means renaming the `.ps1` file to `.psm1`. PowerShell `.psd1` modules are also supported.
* Functions must be exported from the PowerShell module. Note: any function defined in the module scope is exported automatically, this means that no change may be needed for some extensions. However, functions in the global scope are _not_ exported from the module, remove `global:` from any function definitions to ensure they are exported
* Exported functions, MainMenuItems and GameMenuItems functions must accept the exact number of arguments. Previously this was not an error, but now it is.
* PowerShell extensions now run in the STA apartment state instead of MTA.

I have verified that:


bburky:
I was not able to set the MTA apartement state:
```c#
initialSessionState.ApartmentState = System.Threading.ApartmentState.MTA;
```

If I did so, I got this error:
```
When the runspace is set to use the current thread, the apartment state in the invocation settings must match that of the current thread.
```

source/Playnite/Scripting/PowerShell/PowerShell.cs
```diff
@@ -128,12 +138,42 @@ public object GetVariable(string name)
...
+            powershell.Commands.Clear();
+            return command;
+        }
+
+        public object InvokeFunction(string name, List<object> arguments)
+        {
+            var command = GetFunction(name);
+            powershell.AddCommand(command);
+            // Don't pass arguments to commands not expecting parameters
+            if (command.Parameters.Count != 0)
```
bburky, line 159: This is the one change I'm most concerned with breaking compatibility.

Explicitly invoking commands with this API actually enforces that the number of arguments match the command being invoked. That is, providing 2 arguments to a function expecting 0 will cause an error. This is contrary to how PowerShell usually behaves silently ignores the extra arguments.

This is a problem because many extensions ignore arguments passed to them that they don't need. As a workaround, I decided to special case commands expecting exactly 0 arguments and don't pass any if so.

source/Playnite/Scripting/PowerShell/PowerShell.cs
```diff
@@ -50,6 +48,18 @@ public void Dispose()
             runspace.Dispose();
         }
 
+        public void ImportModule(string path)
+        {
+            powershell.Runspace.SessionStateProxy.Path.SetLocation(Path.GetDirectoryName(path));
```
bburky, line 53: I guess setting the working directory of extensions is a breaking change too. I think they previously ran in the Playnite application directory. Maybe this can be removed, I haven't tested without it.

bburky:
Debugging in VS Code works when setting the Integrated console to "Switch to" Windows PowerShell (x64)" to use the built in Windows Powershell 5.1.

bburky:
The [`system.management.automation.powershell` appears to exist in PowerShell SDK 7.0](https://docs.microsoft.com/en-us/dotnet/api/system.management.automation.powershell?view=powershellsdk-7.0.0), which I hope means that this should be easy enough to use in .net core?

JosefNemec:
As far as I know there are no breaking changes in PS Core that would affect Playnite. Other then the fact that the PS language itself has some changes. The integration into Playnite should be as easy as switching to a new nuget package (and of cource the whole Playnite change to .NET 5 before that).

JosefNemec:
I just want to say that I haven't forgot about this. I'll make one or two updates for 8.x first and then immediately start working on Playnite 9 and this is going to be first thing to merge.

JosefNemec:
Can you please re-target this to `devel9` branch and fix the merge conflict in ExtensionFactory? Thanks

bburky:
I dropped commit 13b23b9984891c446daef5936e76eabcf39db274 which made rebasing this easier.

13b23b9984891c446daef5936e76eabcf39db274 injected the `$PlayniteAPI` variable earlier, which isn't required. It would be kind of nice to do, but it would make more sense to lift all variable declarations to be injected earlier. The main reason it would be nice is to allow using the variables outside method definitions.

JosefNemec (commented):
- Did you test impact of these changed on startup times? Is it faster or slower and by how much? Especially related to `GetFunction` changes.
- Did you test that game scripts still work with this change?
- LibraryExporter is not updated to these changes.
- Please update documentation to reflect these changes.

JosefNemec:
One more thing related to detection if extension supports certain functions. If I remember correctly, PS modules should specify list of exported function/cmdlets, can't we use that list?

source/Playnite/Scripting/PowerShell/PowerShell.cs
bburky: I was trying to remove any time strings were eval'ed. Previously the function name and it's args were executed as a string. Maybe there's another API that avoids the parameters problem, but I didn't find it.

bburky:
> Did you test impact of these changed on startup times? Is it faster or slower and by how much? Especially related to GetFunction changes.
No, do you have a good way to measure this?
> LibraryExporter is not updated to these changes.
LibraryExporter works as is with these changes, it is largely backwards compatible. What changes do you suggest?
> Did you test that game scripts still work with this change?
Yes, they were unchanged. I think they are still executed as strings, not modules.

bburky:
>One more thing related to detection if extension supports certain functions. If I remember correctly, PS modules should specify list of exported function/cmdlets, can't we use that list?


I did that in #924, but it is a significantly more breaking change. We must rename extension files to .psm1 and change all functions to not be `global:foo` which is common in existing scripts. With .ps1 files, ExportedFunctions is empty and with .psm1 files it does not include functions in the global scope.

The changes I made were backwards compatible with existing extensions. If you want, we could try supporting both .ps1 and .psm1 files. Then deprecate .ps1 extensions and eventually switch over.

source/Playnite/Scripting/PowerShell/PowerShell.cs
bburky: This runs fine with these changes:
```powershell
$newGame = New-Object "Playnite.SDK.Models.Game"
$newGame.Name = "New Game"
$PlayniteApi.Database.Games.Add($newGame)
```

JosefNemec:
I think we should go full proper PS module support, no half-baked solution. Which yes, would include requirement to have .psm1 (probably also .psd1) files. We are already doing breaking things so it imho doesn't matter. Better do it now, then later find that something else won't work without proper module support and we will be doing another breaking release.

We'll just have to properly document how to convert existing extensions, to make it easier for people (or maybe even write some converter).

> No, do you have a good way to measure this?

I have `ExecutionTimer` class that I use to measure stuff. You can use it to measure how long does it take to load all extensions and compare it with current solution (for example LoadCollections() from GameDatabase uses it to measure how long does it take to load library).

Also, please measure if there's a difference to how long does it take to open game and main menu, because those execute script methods as well (every time they are opened).

bburky:
Ok, I will work to change this to use `ExportedFunctions` instead of using `GetCommand`. Would you like me to make it backwards compatible by falling back to `GetCommand`, or make this a breaking change?

JosefNemec:
I would not add any backwards compatibility, lets do clean implementation using modules.

bburky:
Please see b785d2bf6f1187d52819b95f7db0985a662b138f, I updated everything to use `.psm1` or `.psd1` files.

PowerShell also supports [binary modules with a .dll extension](https://docs.microsoft.com/en-us/powershell/scripting/developer/module/how-to-write-a-powershell-binary-module?view=powershell-7), but I decided supporting these doesn't make any sense. If someone wants this they should write a proper C# Playnite extension.

I changed one reference in the docs to `.psm1` but I probably still need to do more docs updates. I changed all example extensions in the repo to `.psm1` and made functions non-global (I only tested LibraryExporter though). I changed [`Playnite.Toolbox/Templates/Extensions/PowerShellScript/PowerShellScript.psm1`](https://github.com/JosefNemec/Playnite/blob/b785d2bf6f1187d52819b95f7db0985a662b138f/source/Tools/Playnite.Toolbox/Templates/Extensions/PowerShellScript/PowerShellScript.psm1) to have some debugging calls to `$__logger` which I were useful for testing and should be generally useful.

I have not tested speed differences with `ExecutionTimer`.

I tested with both `.psm1` files and `.psd1` files. Migration from `.ps1` to `.psm1` is easy: rename the extension to `.psm1` and update the yaml, then change the exported functions are _not_ be in the global namespace. I didn't use [`Export-ModuleMember`](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/export-modulemember?view=powershell-7#description) because the default behavior works pretty well for Playnite extensions: "If a script module does not include an Export-ModuleMember command, the functions and aliases in the script module are exported, but the variables are not."

When using `.psd1` files you need to place the `.psm1` and `.psd1` file in the same directory, then reference the `.psm1` from the `RootModule` field of the `.psd1` file.
```powershell
@{
RootModule = 'asdf.psm1'
ModuleVersion = '1.0'
GUID = '665d46fb-d343-4c98-8b5c-3869d2d3e0be'
FunctionsToExport = @(
    "OnApplicationStarted",
    "OnApplicationStopped",
    "OnLibraryUpdated",
    "OnGameStarting",
    "OnGameStarted",
    "OnGameStopped",
    "OnGameInstalled",
    "OnGameUninstalled",
    "OnGameSelected"
)
}
```

source/Playnite/Scripting/PowerShell/PowerShell.cs
```diff
@@ -140,15 +140,12 @@ public object GetVariable(string name)
...
-                .Invoke<CommandInfo>().FirstOrDefault();
-            powershell.Streams.ClearStreams();
-            powershell.Commands.Clear();
-            return command;
+            if (module == null)
+            {
+                return null;
+            }
+            CommandInfo command;
+            return module.ExportedCommands.TryGetValue(name, out command) ? command : null;
```
bburky, line 148: I decided to use [`ExportedCommands`](https://docs.microsoft.com/en-us/dotnet/api/system.management.automation.psmoduleinfo.exportedcommands?view=powershellsdk-7.0.0#System_Management_Automation_PSModuleInfo_ExportedCommands) instead of [`ExportedFunctions`](https://docs.microsoft.com/en-us/dotnet/api/system.management.automation.psmoduleinfo.exportedfunctions?view=powershellsdk-7.0.0#System_Management_Automation_PSModuleInfo_ExportedFunctions). `ExportedCommands` appears to be a superset of `ExportedFunctions`. I think this will capture someone defining something as a cmdlet too and not just functions. All the functions do correctly appear in this list also

source/Tests/Playnite.Tests/Scripting/PowerShell/PowerShellTests.cs
```diff
@@ -68,18 +68,23 @@ public void ErrorHandlingTest()
...
+                    Assert.IsTrue(ps.GetFunction("TestFunc") == null);
+                    var path = Path.Combine(tempDir.TempPath, "GetFunctionTest.psm1");
+                    File.WriteAllText(path, @"
 function TestFunc()
 {
     return 4 + 4
 }
 ");
-                Assert.IsTrue(ps.GetFunction("TestFunc") != null);
+                    ps.ImportModule(path);
```
bburky, line 85: I modified this test to create a `.psm1` file and import it as a module. Exported functions will _only_ be found if a module import is done. If you only eval a script and define functions the exported functions will not be detected by `PowerShellRuntime`.

source/Playnite/Scripting/PowerShell/PowerShell.cs
bburky: Would you like me to remove this `if (command.Parameters.Count != 0)` test and require that exported functions accept the correct number of arguments?

source/Playnite/Scripting/PlayniteScript.cs
```diff
@@ -92,8 +92,10 @@ public static PlayniteScript FromFile(string path)
             {
                 return new IronPythonScript(path);
             }
-            else if (extension == ".ps1")
+            // ps1 and dll PowerShell modules are not supported
+            else if (extension == ".psm1" || extension == ".psd1")
```
JosefNemec, line 96: Can't we completely remove module file requirement from PowerShell extensions? As far as I know you should be able to import module from a directory. At least that's how Import-Module used to work in PowerShell 2. That would also add support for compiled modules.

source/Playnite/Scripting/PowerShell/PowerShell.cs
JosefNemec: Yes remove it. Exported functions must have correct number of arguments. This should not be an issue in future because I'll be wrapping all arguments under single object (for better future proofing).

JosefNemec (commented):
> I have not tested speed differences with ExecutionTimer.

I really need to know how big effect these changes have on performance. This is especially critical for game menus and main menu where we do calls to runtime (and extensions) every time menu is opened. I will not merge this until we know if this has negative (and how much) impact on performance.

source/Playnite/Scripting/PlayniteScript.cs
bburky: I can look into loading as a directory, that sou ds good. But the problem is extension loading currently detects native dll/IronPython/PowerShell based off the file extension. I can't remove this test here because it's needed to detect if it's PowerShell at all.

bburky: Also, it looks like using a directory would import _all_ modules in the specified path. The way I wrote this, I need to get a single module as a result and I use it to get it's `ExportedCommands`. I don't think it's too unreasonable for developers to use multiple modules, they may want to bundle an existing PowerShell module as a library they want to use. I don't think the path import works.

JosefNemec: The directory import should import module based on manifest file, if you have some other modules there, then those should be imported only if the manifest for main module imports them as well. I specifically remember using it that way years ago. But again, that was in PS 2.0 days, maybe they changed it.

I only marked this line because I saw specific mention of psm1 file. Of course the module would have to be kept in our manifest for Python extensions, but it can be removed for PS ones. If it's specified and is .py, then it's Python, otherwise treat it as PS.

bburky:
I added `ExecutionTimer` instrumentation. The new implementation appears to be slightly 
faster actually. See [this playnite.log](https://github.com/JosefNemec/Playnite/files/5490563/playnite.log).

In both the new version and the old version the very first `PlayniteScript.FromFile` call is very slow. I'm not sure if this is due to other high CPU usage during Playnite startup impacting it, or some actual PowerShell startup. Every extension after that (and reloading extensions) takes ~50-150ms. The old implantation also had the same first time slowness, but other extensions load in ~150-200 ms. Basically the same speed.

The first call to `script.GetMainMenuItems` is the same in both the old and new implementation. But successive calls appear to take 0-2ms in the new implementation vs ~5ms with the old implementation. Basically the same speed, but maybe a tiny bit faster.

I used the LibraryExporter and PowerShellScript extensions for the above tests. Neither of those had a `GetGameMenuItems` so I didn't test it. I can if you'd like but I expect it to be the same as `GetMainMenuItems`

If you'd like to test this yourself you can cherry pick the `ExecutionTimer` commit to test with the old implementation:
```
git checkout devel9 -b old-powershell-with-ExecutionTimer
git cherry-pick 717020ba9626847e60fb86af871443bf5b6db400
```

source/Playnite/Scripting/PowerShell/PowerShell.cs
```diff
@@ -28,19 +32,13 @@ public static bool IsInstalled
 
         public PowerShellRuntime(string runspaceName = "PowerShell")
         {
-            runspace = RunspaceFactory.CreateRunspace();
+            initialSessionState = InitialSessionState.CreateDefault();
+            initialSessionState.ExecutionPolicy = ExecutionPolicy.Unrestricted;
+            initialSessionState.ThreadOptions = PSThreadOptions.UseCurrentThread;
```
bburky, line 37: I tried to figure out how to get `System.Management.Automation.PowerShell` to work with an MTA apartment state, but couldn't get it to work well. As I mentioned in a previous comment, I get this error if I set it:
>When the runspace is set to use the current thread, the apartment state in the invocation settings must match that of the current thread.

The following allows MTA to work, but breaks usage of GUI from extensions:
```
initialSessionState.ApartmentState = System.Threading.ApartmentState.MTA;
initialSessionState.ThreadOptions = PSThreadOptions.ReuseThread;
```
LibraryExporter just hangs after clicking its menu item. The file dialog never appears.

I don't know STA vs MTA differences well enough. What are the implications of using STA? Will that break existing extensions?

It might be possible to manually create the runspace like the old code did to get the old behavior back.

bburky: I tried manually creating the runspace like this, but it doesn't work either. I get the same error.
```csharp
initialSessionState = InitialSessionState.CreateDefault();
initialSessionState.ExecutionPolicy = ExecutionPolicy.Unrestricted;

runspace = RunspaceFactory.CreateRunspace(initialSessionState);
runspace.Name = runspaceName; runspace.Name = runspaceName;
runspace.ApartmentState = System.Threading.ApartmentState.MTA;
runspace.ThreadOptions = PSThreadOptions.UseCurrentThread;
runspace.Open();

powershell = System.Management.Automation.PowerShell.Create();
powershell.Runspace = runspace;
```

JosefNemec: We have to use STA because of GUI interaction, there's no workaround as far as I know.

JosefNemec:
The very first call always takes longer because PowerShell is initializing for the first time (loading assemblies into process etc.).

source/Playnite/Scripting/PowerShell/PowerShell.cs
JosefNemec: I think the only solution would be to create separate thread for PS extensions and then re-implementing every GUI call to be thread safe. But I don't think that's worth the trouble, because Playnite always has to wait for script to finish anyways, so the execution will always act as synchronous not matter what. We should also keep it synchronous to have the same execution behavior consistent with plugins.

bburky: Okay, then we're good. It's already STA. Somehow the old PowerShell implementation was MTA and had working GUI though. Perhaps it's the new method of invoking PowerShell I'm using, but it now gives me errors with MTA+UseCurrentThread. As long as STA isn't a breaking change, I'm good with this.

JosefNemec: Yeah, it's because you now execute methods differently. I have sneaking suspicion that we might need to change this once we switch to PS Core (because we will ship Playnite with built-in PS runtime), but I haven't tested it yet.

Now that I'm thinking about it, this could be an issue for game scripts, because those are still being executed the old way, right? But I don't know how likely is that someone will be using GUI code there though.

doc/tutorials/menus.md
```diff
@@ -13,6 +13,8 @@ To add new custom menu entries, implement appropriate menu script/plugin functio
 
 You can use those argument objects to decide what elements to return. For example, in case of game menu, `GetGameMenuItemsArgs` contains `Games` field listing all currently selected games that are being used as a source for the game menu.
 
+When the function associated with the menu item is invoked, it will be passed an arguments object. For game menu actions, C# extensions will be passed [GameMenuItemActionArgs](xref:Playnite.SDK.Plugins.GameMenuItemActionArgs) and script extensions will be passed [ScriptGameMenuItemActionArgs](xref:Playnite.SDK.Plugins.ScriptGameMenuItemActionArgs). `GameMenuItemActionArgs` has a `Games` property to access the selected list of games. For main menu actions, C# extensions will be passed [MainMenuItemActionArgs](xref:Playnite.SDK.Plugins.MainMenuItemActionArgs) and script extensions will be passed [ScriptMainMenuItemActionArgs](xref:Playnite.SDK.Plugins.ScriptMainMenuItemActionArgs).
```
bburky, line 16: I think this is correct, I don't fully understand the difference between `GameMenuItemActionArgs` and `ScriptGameMenuItemActionArgs`.

doc/tutorials/scripts/scripting.md
```diff
@@ -9,6 +9,11 @@ Playnite can be extended with additional functionality using scripts. [PowerShel
...
+
+> [!NOTE]
+> Importing PowerShell extensions as modules is a breaking change as of Playnite 9. Existing extensions will be need to be updated. For many extensions this will be as simple as renaming the file from `.ps1` to `.psm1`, updating the manifest and removing `global:` from any exported functions. A `params()` section may need to be added to exported functions that were not explicitly accepting the correct number of arguments.
```
bburky, line 15: I think this and the note above it are enough to explain the new changes about importing PowerShell scripts as modules. Please let me know if this needs to be clarified or expanded anywhere.

doc/tutorials/windows.md
```diff
@@ -32,4 +32,65 @@ window.WindowStartupLocation = WindowStartupLocation.CenterOwner;
...
+        </ItemsControl>
+        <StackPanel Orientation="Horizontal" Margin="0,15,0,0">
+            <TextBlock Text="Total selected games: "/>
+            <TextBlock Text="{Binding .Count}"/>
+        </StackPanel>
+    </StackPanel>
+</UserControl>
+"@
+$reader = [System.Xml.XmlNodeReader]::new($xaml)
+$window.Content = [Windows.Markup.XamlReader]::Load($reader)
```
bburky, line 79: I added a new PowerShell example using `PlayniteApi.Dialogs.CreateWindow()` to demonstrate this API. I used my old XAML test code adapted to use a `UserControl` as the example.

I like using an XAML example like this to demonstrate this feature, but I don't mind if we tweak the exact contents of this sample XAML. I think a complete example should probably be added to the C# tab too.

JosefNemec:
I've made some major refactoring of documentation and forgot to merge it into devel9, sorry about that.

doc/tutorials/scripts/scripting.md
JosefNemec: I think it would be better to have it in the changelog and create separate page that explains how to migrate from Playnite 8 to 9. People just starting with scripting won't care about breaking changes from earlier versions.

doc/tutorials/menus.md
JosefNemec: One is passed when menu items are requested and the other one passed when menu item is invoked.

bburky: No, what I was referring to was the script variant of the same classes.

ScriptMainMenuItemActionArgs and ScriptMainMenuItemActionArgs
vs
GetMainMenuItemsArgs and MainMenuItemActionArgs

JosefNemec: One set is for scripts and the other is for compiled plugins.

bburky: Yeah, that's what I thought. Just making sure I understood it correctly. That's what's in the text I added to the docs here.

doc/tutorials/scripts/scripting.md
bburky: Ok, I added the note as a **Breaking Change** in the changelog. I put it in a new "TODO: 5.x.x" section, you can update it when you make a release. I left it as paragraph text, not bullet points. If you want it restructured we can change it to bullet points.

bburky:
Alright, I think I successfully merged in the docs changes.

source/Playnite/Scripting/PlayniteScript.cs
bburky: Okay, it does appear to do the right thing and only import one module, it doesn't import all modules like I was concerned about. However, this _requires_ the name of the module to match the name of the directory. I guess developers can control the name of the extension and this would work, but I think this could be confusing.

```
mkdir moduleTest
echo 'function foo() { echo "foo" }' > moduleTest\foo.psm1
echo 'function bar() { echo "bar" }' > moduleTest\moduleTest.psm1
$module = Import-Module -PassThru .\moduleTest
# Module is from moduleTest.psm1, and foo.psm1 is not imported
```

JosefNemec:
Last thing, can you remove those ExecutionTimer uses you added (I'll be adding more universal profiling via #2093)?

PR #2126, opened 2020-11-21
Refernce `pwsh.exe` without explicit path

tylerszabo:
PowerShell installs into different directories depending on how it was
installed. Use whatever `pwsh.exe` is in the path but explicitly check
version using `#Requires -Version` in the script files.

This change targets `master` because it's a fix for a build break rather than new feature work.

I have verified that:


JosefNemec:
This step has been completely removed from build process in Playnite 9 (`devel9`).

PR #2132, opened 2020-11-22
Add xenia canary support

greggameplayer:
I have verified that:
- [X] I added myself into [contributors file]

Modify regex to also look at Xenia canary binary.

### DIFF
--- a/source/Playnite.DesktopApp/Resources/contributors.txt
+++ b/source/Playnite.DesktopApp/Resources/contributors.txt
@@ -40,3 +40,4 @@ Tyler Szabo
 elisherer
 awbooze
 erri120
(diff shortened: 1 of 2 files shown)


darklinkpower:
In my opinion each version should be kept in separate profiles in case someone wants or needs to have the two versions, like the PPSSPP profile for example: https://github.com/JosefNemec/Playnite/blob/master/source/Playnite/Emulators/PPSSPP.yaml

But it's up to @JosefNemec to decide that.

JosefNemec:
I agree. They should be separate profiles.

greggameplayer:
@JosefNemec done

Issue #2215, opened 2021-01-08
Importing single .lnk discards start arguments


felixkmh:
**Description**
When using _Add Game_ -> _Scan automatically_ -> _Browse..._ to add a Game via its shortcut (.lnk), its launch arguments are discarded as not the .lnk gets imported, but the .exe it points to because the file selection dialog filters for .exe files.

So to fix that, the filter also needs to accept .lnk files, similar to what happens when importing a whole folder automatically. And I would suggest using the .lnks file name as the game name in that case.

**Reproduction**
Use   _Add Game_ -> _Scan automatically_ -> _Browse..._ to import a game using its .lnk shortcut. E.g. all games using RiotClient.exe with arguments to launch different games.

felixkmh:
fixed via #2216

PR #2216, opened 2021-01-08
Improved single file import in InstalledGamesView for .lnk files

felixkmh:
addressing #2215 
import single files similarly to how a whole folder is imported. Addtionally enables to import single .lnk files pointing to .exe or .bat files, or .bat files themselves.
Also: for .lnk files the file name of the shortcut is used as the game name, if it is not empty.

### DIFF
--- a/source/Playnite.DesktopApp/ViewModels/InstalledGamesViewModel.cs
+++ b/source/Playnite.DesktopApp/ViewModels/InstalledGamesViewModel.cs
@@ -340,23 +340,34 @@ namespace Playnite.DesktopApp.ViewModels
 
         public void SelectExecutable()
         {
-            var path = dialogs.SelectFile("Executable (.exe)|*.exe");
+            var path = dialogs.SelectFile("Executable (.exe,.bat,lnk)|*.exe;*.bat;*.lnk");
             if (string.IsNullOrEmpty(path))
             {
                 return;
             }
+            var program = Common.Programs.GetProgramData(path);
 
-            var productName = FileVersionInfo.GetVersionInfo(path).ProductName;
-            var import = new ImportableProgram(new Program()
+            // Ignore all target files that are not bat or exe files
+            if (!program.Path.EndsWith(".exe", StringComparison.OrdinalIgnoreCase) &&
+                !program.Path.EndsWith(".bat", StringComparison.OrdinalIgnoreCase))
             {
-                Icon = path,
-                Name = string.IsNullOrWhiteSpace(productName) ? new DirectoryInfo(Path.GetDirectoryName(path)).Name : productName,
-                Path = path,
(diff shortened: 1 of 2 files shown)


source/Playnite.DesktopApp/ViewModels/InstalledGamesViewModel.cs
```diff
@@ -340,24 +340,35 @@ public void ConfirmDialog()
...
+                return;
+            }
+
+            var import = new ImportableProgram(program, ProgramType.Win32)
             {
                 Selected = true
             };
 
+            // Use shortcut name as game name for .lnk shortcuts
+            if (Path.GetExtension(path).EndsWith(".lnk", StringComparison.OrdinalIgnoreCase))
```
JosefNemec, line 363: Calling GetExtension is not necessary since you are checking with EndsWith anyways.

source/Playnite.DesktopApp/ViewModels/InstalledGamesViewModel.cs
```diff
@@ -340,24 +340,35 @@ public void ConfirmDialog()
...
             if (string.IsNullOrEmpty(path))
             {
                 return;
             }
+            var program = Common.Programs.GetProgramData(path);
 
-            var productName = FileVersionInfo.GetVersionInfo(path).ProductName;
-            var import = new ImportableProgram(new Program()
+            // Ignore all target files that are not bat or exe files
+            if (!program.Path.EndsWith(".exe", StringComparison.OrdinalIgnoreCase) &&
```
JosefNemec, line 351: I don't think we should have this restriction. If somebody decides to import .lnk, then we should not care what that .lnk points to.

source/Playnite.DesktopApp/ViewModels/InstalledGamesViewModel.cs
felixkmh: resolved in https://github.com/felixkmh/Playnite/commit/f5f7cfdfadfc60fcadb8493435646768b61e68f1

source/Playnite.DesktopApp/ViewModels/InstalledGamesViewModel.cs
felixkmh: resolved in https://github.com/felixkmh/Playnite/commit/8ac8f5f57d286b3a1a80bdd23bbe0e0df33ad885

felixkmh: I think the rationale behind this was that you previously couldn't select anything but .exe and .bat files in the first place. The only reason you could select .lnk files was that the file selection dialogue returned the path to the file the shortcut pointed to, which then had to match the filter.

PR #2217, opened 2021-01-09
SDK Feature: Select Game

felixkmh:
Exposed the SelectGame(Id) method from DesktopViewModels to the MainViewApi. Something I would like to use for my DuplicateHider extension so users can select/view copies of games without triggering their installation or launch.


JosefNemec:
Added with some other changes via b26a83fd936c24689144d26a79323740828cb245

PR #2256, opened 2021-02-12
Take 2: Added emulators.

wrennnnnn:
I have verified that:
- [Yes] These changes work, by building the application and testing them.
- [Yes] Pull request is targeting `devel` branch.
- [No] I added myself into [contributors file](https://github.com/JosefNemec/Playnite/blob/devel/source/Playnite.DesktopApp/Resources/contributors.txt) if I want to receive contribution credit in the application itself.
Sorry about that. Added support for Flycast and FBNeo, as well as updating the Fightcade handler.

### DIFF
--- a/source/Playnite/Emulators/FightCade.yaml
+++ b/source/Playnite/Emulators/FightCade.yaml
@@ -5,4 +5,4 @@ Profiles:
     DefaultArguments: '"{ImageNameNoExt}"'
     Platforms: [Various]
     ImageExtensions: [zip, bin]
-    ExecutableLookup: ^ggpofba\.exe$
+    ExecutableLookup: ^fcadefbneo\.exe$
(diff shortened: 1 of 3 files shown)


source/Playnite/Emulators/FightCade.yaml
```diff
@@ -5,4 +5,4 @@ Profiles:
     DefaultArguments: '"{ImageNameNoExt}"'
     Platforms: [Various]
     ImageExtensions: [zip, bin]
-    ExecutableLookup: ^ggpofba\.exe$
```
JosefNemec, line 8: Did this permanently change in the app? Won't this prevent use of older versions?

wrennnnnn:
It permanently changed (with the release of Fightcade 2.) It also auto
updates on launch and is unplayable without that automatic update. I can’t
see any use case in which you have a working old version of Fightcade.

PR #2324, opened 2021-05-06
Exit fullscreen

yeahme49:
This allows you to exit fullscreen by pressing B while on the main screen. It's asks for confirmation before exiting

I have verified that:


JosefNemec:
1) You have 136 file changed in this PR, which is obviously wrong.
2) You are using tabs instead of spaces which is against code contribution rules.
3) I'm personally not a big fan of this. It adds new button to already crowded view (bottom of the main view) and I don't think that just pressing B from main screen should initiate application shutdown. It's imho not a good UX.

PR #2325, opened 2021-05-06
Exit fullscreen

yeahme49:
I have verified that:

I closed the previous PR due to the excess commits. This now just contains my change. I also fixed the tabs.

This commit doesn't just initiate application shutdown when you press B, it asks for confirmation before it closes. I find this to be useful because if I'm using a controller for everything, I don't want to get my mouse and keyboard to exit Playnite. My use-case is launching Playnite from Kodi to play games and then back to Kodi for TV watching on my HTPC. Not having a way to exit with the controller adds extra steps. I know I'm not the only one that wants this, I saw a request on the subreddit asking for the same thing.

If the UI is too crowded at the bottom, the indicator for exiting could be removed but still have the functionality. Would it be better to have a setting option to enable/disable exiting by pressing B?


JosefNemec:
I still maintain my opinion from original PR. Also, you don't need KB and mouse to shutdown FS mode, you can open main menu with a controller and shutdown from there.

yeahme49:
Lots of games allow exiting to desktop with the push of a controller button, typically they confirm is that's what you want. How is this any different?

I would agree with you if pressing B just instantly closed the application without confirmation, but it doesn't.

JosefNemec:
> How is this any different?

Playnite is not a game.

PR #2375, opened 2021-07-08
New: Epic playtime import support

darklinkpower:
Adds playtime import to the Epic library plugin. Based on research done by joyrider3774.

I have verified that:

### DIFF
--- a/source/Plugins/EpicLibrary/Services/EpicAccountClient.cs
+++ b/source/Plugins/EpicLibrary/Services/EpicAccountClient.cs
@@ -49,7 +49,7 @@ namespace EpicLibrary.Services
             var accountUrlMask = @"https://{0}/account/api/public/account/";
             var assetsUrlMask = @"https://{0}/launcher/api/public/assets/Windows?label=Live";
             var catalogUrlMask = @"https://{0}/catalog/api/shared/namespace/";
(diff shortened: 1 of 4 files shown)


source/Plugins/EpicLibrary/Services/EpicAccountClient.cs
```diff
@@ -233,6 +247,11 @@ private void renewToknes(string refreshToken)
         {
             using (var httpClient = new HttpClient())
             {
+                if (url.Contains("/library/api/public/playtime/account/"))
```
JosefNemec, line 250: I don't think this should be placed here. Generic method like this should not be doing arbitrary modifications to passed parameters. Url modifications should be done in GetPlaytimeItems.

darklinkpower: I agree. I was not sure if this was a good idea for that reason but decided this initially to not have to load the tokens twice.

darklinkpower: Updated PR with the change

JosefNemec:
Merged manually into new extensions repo https://github.com/JosefNemec/PlayniteExtensions/commit/9cbc78452cd4acd0515d459344e4bad587fe2772

PR #2385, opened 2021-07-16
Fix: Prevent updating game library entries during library update when no data has changed. Fixes #2384

### LINKED ISSUE #2384: Prevent Playnite from updating game database file when new data is not different on library update
**Describe the bug**
Updating a library will cause Playnite to update all game database files, even when the new data has no changes from the current data.

Currently, it has been confirmed that the `IsInstalled` property is updated everytime even when there is no change.

**To Reproduce**
Update any game library.

darklinkpower:
Fix: Prevent updating game library entries during library update when no data has changed. Fixes #2384

I have verified that:

### DIFF
--- a/source/Playnite/Database/GameDatabase.cs
+++ b/source/Playnite/Database/GameDatabase.cs
@@ -1065,8 +1065,8 @@ namespace Playnite.Database
                             existingGame.IsInstalled = newGame.IsInstalled;
                             existingGameUpdated = true;
                         }
-                        
-                        if (existingGame.InstallDirectory.Equals(newGame.InstallDirectory, StringComparison.OrdinalIgnoreCase) == false)
+
+                        if (string.Equals(existingGame.InstallDirectory, newGame.InstallDirectory, StringComparison.OrdinalIgnoreCase) == false)
                         {
                             existingGame.InstallDirectory = newGame.InstallDirectory;
                             existingGameUpdated = true;
--- a/source/Playnite/Database/GameDatabase.cs
+++ b/source/Playnite/Database/GameDatabase.cs
@@ -1059,34 +1059,61 @@ namespace Playnite.Database
                     }
                     else
                     {
-                        existingGame.IsInstalled = newGame.IsInstalled;
-                        existingGame.InstallDirectory = newGame.InstallDirectory;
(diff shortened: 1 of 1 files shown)


source/Playnite/Database/GameDatabase.cs
```diff
@@ -1059,34 +1059,60 @@ public List<Game> ImportGames(LibraryPlugin library, bool forcePlayTimeSync, ILi
...
-                        existingGame.IsInstalled = newGame.IsInstalled;
-                        existingGame.InstallDirectory = newGame.InstallDirectory;
+                        var existingGameUpdated = false;
+                        if (existingGame.IsInstalled != newGame.IsInstalled)
+                        {
+                            existingGame.IsInstalled = newGame.IsInstalled;
+                            existingGameUpdated = true;
+                        }
+                        
+                        if (existingGame.InstallDirectory.Equals(newGame.InstallDirectory, StringComparison.OrdinalIgnoreCase) == false)
```
JosefNemec, line 1069: This will throw if `InstallDirectory` is null.

source/Playnite/Database/GameDatabase.cs
```diff
@@ -1059,34 +1059,60 @@ public List<Game> ImportGames(LibraryPlugin library, bool forcePlayTimeSync, ILi
...
+                        {
+                            existingGame.InstallDirectory = newGame.InstallDirectory;
+                            existingGameUpdated = true;
+                        }
+                        
                         if (existingGame.PlayAction == null || existingGame.PlayAction.IsHandledByPlugin)
                         {
-                            existingGame.PlayAction = newGame.PlayAction;
+                            if (existingGame.PlayAction != newGame.PlayAction)
```
JosefNemec, line 1077: GameAction object doesn't implement != operator, so this check will not work and always will be false.

source/Playnite/Database/GameDatabase.cs
darklinkpower: Fixed by using `string.Equals` method instead.

darklinkpower:
Only remaining issue is the check of the `PlayAction` but as discussed on discord will be ignored for now since they are handled differently in Playnite 9.

PR #2438, opened 2021-09-04
Devel: add linux and mac platforms

grnassar:
I have verified that:

Edits to use new PN9 MetadataSpecProperty-based platforms listing throughout:
- change an instance of MetadataName to MetadataSpec
- add mac and linux platforms (naming and ID taken from IGDB's platforms info)


source/Tools/Playnite.Toolbox/Playnite.Toolbox.csproj
```diff
@@ -71,10 +71,6 @@
       <Link>Templates\Themes\Fonts\icofont.ttf</Link>
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
-    <None Include="..\..\..\references\Fonts\PlayStation4.ttf">
-      <Link>Templates\Themes\Fullscreen\Fonts\PlayStation4.ttf</Link>
```
JosefNemec, line 75: Really? Just removing files from a project without any comment or anything?

source/Playnite/Emulation/Platforms.yaml
```diff
@@ -376,6 +376,16 @@
   IgdbId: 6
   Emulators: [pcem, scummvm]
 
+- Name: Linux
```
JosefNemec, line 379: This should be imho `PC (Linux)` for consistency and mac should be `Macintosh`.

source/Tools/Playnite.Toolbox/Playnite.Toolbox.csproj
grnassar: Oof. My bad. Actually, *you* deleted the file! Check your repo -- the files don't exist. It just hadn't been removed from the .csproj, and the fact that it still shows up in .csproj breaks any build done from scratch.  So either the file needs to be readded or the .csproj does need to be changed.  But *neither* one of those obviously belongs in this pull request. I thought I'd branched off those changes separately but apparently I didn't.

source/Playnite/Emulation/Platforms.yaml
grnassar: I'm fine with that.  I just used the naming that IGDB used, as I thought that's what you were going for.  It uses "PC (Windows)" and "PC (DOS)", but "Linux" and "Mac" rather than "PC (Linux)" and "Macintosh".  Those are obviously all just arbitrary choices, so it's all the same to me as long as the types themselves are included.

Due to the error I caught above (re: not having submitted this off its own branch, and therefore it including other random stuff) I'll retract this PR and resubmit.  Unless you want to just insert the two types in yourself, certainly fine by me too.

source/Tools/Playnite.Toolbox/Playnite.Toolbox.csproj
JosefNemec: https://github.com/JosefNemec/Playnite/wiki/Building#fullscreen-mode-development

This applies to Toolbox as well since it's handling Fullscreen templates.

source/Playnite/Emulation/Platforms.yaml
JosefNemec: We don't use names from IGDB, so I don't see why we should use it for these two. If some match with what's on IGDB, it's just a coincidence. Most of the naming is based on what libretrodb uses.

grnassar: Sounds good to me.  Like I said, it's all just arbitrary choices.  I had assumed IGDB since the current ones all match it exactly. (Libretro, to my knowledge, doesn't have PC/Windows/Linux/Mac platforms specified *at all*; it does have a DOS platform, but while ours and IGDB's is called "PC (DOS)" libretro's is just called "DOS".)

I should have this updated in a few minutes.

source/Tools/Playnite.Toolbox/Playnite.Toolbox.csproj
grnassar: Link for the resources in that page leads to a page that says:
"Unfortunately, Xbox One & PlayStation 4 Buttons Pack is no longer available."

Perhaps the right solution, then, is to take Toolbox out of the default build and let folks enable it if they need it.  Probably beats breaking a default build from a fresh Git pull, I'd think.  (And the dependency of Toolbox on those font assets should also be mentioned in the build doc; Toolbox is never mentioned in the build doc at all, and I know I haven't even used Toolbox and have never even had the need to find out what it does, so I definitely wouldn't know that it's "handling Fullscreen templates" and wouldn't have any way to directly find out.  In the end, the fact that a from-scratch clean build would break due to a dependency on a dev tool I've never used nor needed to would seem to imply *something* there needs fixing one way or the other.)

Regardless, this has been pulled out of this PR.  It certainly doesn't belong here.

source/Playnite/Emulation/Platforms.yaml
grnassar: Just to make sure: "Macintosh" and not "MacOS" and/or "OS X"?  Again, it's just a personal choice; just wanted to present the option.  (I have a Mac that runs Linux on it, so whatever software I run on it would fall under both platforms "PC (Linux)" and "Macintosh" -- or neither, perhaps? Who knows, this platform stuff is messy no matter what. Hence arbitrary choices) :)

grnassar: Renamed as requested.

JosefNemec: I don't know, it's arbitrary. If we were to go by the current naming, then it should be "macOS" :)

JosefNemec:
Btw this was not tested in the actual application! There's a typo I missed during the review and it prevents platforms to be deserialized properly. Test you changes!!!! There's a reason why we have that step in PR template.

PR #2508, opened 2021-09-28
ListToStringConverter seperator as parameter

felixkmh:
Added custom seperator string to ListToString converter using the parameter object

I have verified that:

### DIFF
--- a/source/Playnite/Converters/ListToStringConverter.cs
+++ b/source/Playnite/Converters/ListToStringConverter.cs
@@ -50,6 +50,8 @@ namespace Playnite.Converters
 
     public class ListToStringConverter : MarkupExtension, IValueConverter
     {
+        private const string defaultSeperator = ",";
+    
         public static string MakeString(IEnumerable<string> source)
         {
             return string.Join(",", source);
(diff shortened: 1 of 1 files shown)


JosefNemec:
Why? Where is this used? Please create test for this.

felixkmh:
It's just a modification I made to use this Converter in one of my extensions, so I could, for example, use either a newline character or a ", " as a seperator to display list data like platforms, developers, genres, etc., without creating a new Converter for every seperator. And because it does not break the original behaviour either, I thought I might as well push it back up.

source/Playnite/Converters/ListToStringConverter.cs
```diff
@@ -62,7 +64,14 @@ public object Convert(object value, Type targetType, object parameter, System.Gl
                 return string.Empty;
             }
 
-            return string.Join(",", (IEnumerable<object>)value);
+            string sep = defaultSeperator;
+
```
JosefNemec, line 68: Remove these empty lines please.

felixkmh: done

PR #2662, opened 2021-11-20
Sorting name prefill

Jeshibu:
Added the ability to auto-fill the Sorting Name field when importing games.
This removes (user-adjustable) articles (default: The/A/An) and converts roman numerals to arabic numerals. All numbers are left-padded to a minimum length of 2 with zeroes, except those that start the string, and "I" that's not at the end of the string. Some common abbreviations like XXL are also excluded from roman numeral parsing. Many of the cases that lead to these exceptions are included in tests.

I have verified that:


source/Playnite.DesktopApp/ViewModels/SettingsViewModel.cs
```diff
@@ -233,6 +233,71 @@ public RelayCommand<object> WindowClosingCommand
...
+                if (res.Result && !res.SelectedString.IsNullOrEmpty())
+                {
+                    if (settings.GameSortingNameRemovedArticles.Any(a => a.Equals(res.SelectedString, StringComparison.InvariantCultureIgnoreCase)))
+                    {
+                        dialogs.ShowErrorMessage(resources.GetString("LOCItemAlreadyExists"), "");
+                    }
+                    else
+                    {
+                        settings.GameSortingNameRemovedArticles.Add(res.SelectedString);
+                        settings.GameSortingNameRemovedArticles = settings.GameSortingNameRemovedArticles.GetClone();
```
JosefNemec, line 253: I would just create backing property that implements observable collection and update source settings when settings are being saved. Like we already do that with `ImportExclusionList` list.

source/Playnite.DesktopApp/ViewModels/SettingsViewModel.cs
```diff
@@ -233,6 +233,71 @@ public RelayCommand<object> WindowClosingCommand
...
+        public RelayCommand<object> FillSortingNameForAllGames
+        {
+            get => new RelayCommand<object>(_ =>
+            {
+                dialogs.ActivateGlobalProgress(args =>
+                {
+                    args.ProgressMaxValue = database.Games.Count;
+                    var updatedGames = new List<SDK.Models.Game>();
+                    var c = new SortableNameConverter(settings.GameSortingNameRemovedArticles, true);
+                    foreach (var game in database.Games)
```
JosefNemec, line 280: This should be done via `BufferedUpdate` (on `database`) to not call game update events for each game individually.

source/Playnite.DesktopApp/ViewModels/SettingsViewModel.cs
```diff
@@ -233,6 +233,71 @@ public RelayCommand<object> WindowClosingCommand
...
+        }
+
+        public RelayCommand<object> FillSortingNameForAllGames
+        {
+            get => new RelayCommand<object>(_ =>
+            {
+                dialogs.ActivateGlobalProgress(args =>
+                {
+                    args.ProgressMaxValue = database.Games.Count;
+                    var updatedGames = new List<SDK.Models.Game>();
```
JosefNemec, line 278: This is not used anywhere.

source/Playnite.DesktopApp/ViewModels/SettingsViewModel.cs
```diff
@@ -233,6 +233,71 @@ public RelayCommand<object> WindowClosingCommand
...
+                            string sortingName = c.Convert(game.Name);
+                            if (game.Name != sortingName)
+                            {
+                                game.SortingName = sortingName;
+                                database.Games.Update(game);
+                            }
+                        }
+                        args.CurrentProgressValue++;
+                    }
+                }, new GlobalProgressOptions(resources.GetString("LOCSortingNameAutofillProgress"), true));
```
JosefNemec, line 297: Use LOC constant.

source/Playnite/Localization/LocSource.xaml
```diff
@@ -1022,6 +1022,13 @@ If you start experiencing some issues after installing a theme or an extension,
...
+    <sys:String x:Key="LOCSortingNameAutofill">Automatically fill sorting name for batch-added games</sys:String>
+    <sys:String x:Key="LOCSortingNameAutofillTooltip">When you add games via a library update, an emulator folder scan, or a normal folder scan, automatically fill the "Sorting Name" field with a better sortable representation of the game's name, for example "The Witcher 3" will get a Sorting Name of "Witcher 03"</sys:String>
+    <sys:String x:Key="LOCSortingNameRemovedArticles">These words will be removed from the start of the automatically filled Sorting Name value:</sys:String>
+    <sys:String x:Key="LOCSortingNameRemovedArticlesTooltip">Use this for ignoring words at the start of a string for sorting purposes. The default is "The", "An", and "A".</sys:String>
+    <sys:String x:Key="LOCSortingNameAutofillButton">Fill Sorting Name for games without one</sys:String>
+    <sys:String x:Key="LOCSettingsSortingLabel">Sorting</sys:String>
+    <sys:String x:Key="LOCSortingNameAutofillProgress">Filling Sorting Name values...</sys:String>
```
JosefNemec, line 1031: Use ellipsis character instead of `...`.

source/Playnite/SortableNameConverter.cs
```diff
@@ -0,0 +1,209 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+
+namespace Playnite
+{
+    public class SortableNameConverter
+    {
+        public IEnumerable<string> Articles { get; }
```
JosefNemec, line 10: This is not used anywhere.

source/Playnite/SortableNameConverter.cs
```diff
@@ -0,0 +1,209 @@
...
+
+        private Regex regex;
```
JosefNemec, line 12: Should probably be readonly.

source/Playnite/SortableNameConverter.cs
```diff
@@ -0,0 +1,209 @@
...
+
+        /// <summary>
+        /// The minimum string length of numbers. If 4, XXIII or 23 will turn into 0023.
+        /// </summary>
+        private static int NumberLength = 2;
```
JosefNemec, line 17: Wrong naming, private members must use camelCase.

source/Playnite/SortableNameConverter.cs
```diff
@@ -0,0 +1,209 @@
...
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="articles">Words to remove from the start of the title. Suggested: the contents of PlayniteSettings.GameSortingNameRemovedArticles, or "The", "A", "An".</param>
+        /// <param name="batchOperation">Optimize for larger amounts of throughput. Slower for small amounts.</param>
+        public SortableNameConverter(IEnumerable<string> articles, bool batchOperation = false)
+        {
+            Articles = articles ?? throw new ArgumentNullException(nameof(articles));
+            string articlesPattern = string.Join("|", articles.Select(Regex.Escape));
```
JosefNemec, line 32: Use var.

source/Playnite/SortableNameConverter.cs
```diff
@@ -0,0 +1,209 @@
...
+            var options = RegexOptions.ExplicitCapture;
+            if (batchOperation)
+                options |= RegexOptions.Compiled;
```
JosefNemec, line 35: Wrong styling, missing {}.

source/Playnite/SortableNameConverter.cs
```diff
@@ -0,0 +1,209 @@
+﻿using System;
```
JosefNemec, line 1: Please try to follow some sensible ordering of class members. Like all private fields/properties, then public, then constructor then methods. It's mixed all together now.

source/Playnite/SortableNameConverter.cs
```diff
@@ -0,0 +1,209 @@
...
+                {
+                    output -= value;
+                }
+                else
+                {
+                    output += value;
+                    biggestNumberToTheRight = value;
+                }
+
+                #region validation
```
JosefNemec, line 151: This region marking doesn't make sense to me, it's almost the entire for loop in it.

source/Playnite/ViewModels/MainViewModelBase.cs
```diff
@@ -640,18 +640,42 @@ private async Task UpdateLibraryData(Func<CancellationToken, List<Game>> updateA
                         Logger.Info($"Downloading metadata for {addedGames.Count} new games.");
                         ProgressValue = 0;
                         ProgressTotal = addedGames.Count;
-                        ProgressStatus = Resources.GetString(LOC.ProgressMetadata);
+                        string progressBaseStr = ProgressStatus = Resources.GetString(LOC.ProgressMetadata);
```
JosefNemec, line 643: I don't understand why is this done this way now.

source/Playnite.DesktopApp/ViewModels/DesktopAppViewModel.cs
```diff
@@ -593,6 +648,7 @@ public async void ImportEmulatedGames(EmulatedGamesImportViewModel model)
                     Logger.Warn("Skipping metadata download for manually added emulated games, some global task is already in progress.");
                 }
             }
+            await SetSortingNames(model.ImportedGames, true);
```
JosefNemec, line 651: This will cause sorting names to be applied twice when UpdateEmulationLibrary is used via UpdateLibraryData.

source/Playnite/ViewModels/MainViewModelBase.cs
```diff
@@ -640,18 +640,42 @@ private async Task UpdateLibraryData(Func<CancellationToken, List<Game>> updateA
...
                     }
+
+                    if (addedGames.Any() && AppSettings.GameSortingNameAutofill)
+                    {
+                        Logger.Info($"Setting Sorting Name for {addedGames.Count} new games.");
+                        ProgressValue = 0;
+                        ProgressTotal = addedGames.Count;
+                        string progressBaseStr = ProgressStatus = Resources.GetString(LOC.SortingNameAutofillProgress);
+                        var c = new SortableNameConverter(AppSettings.GameSortingNameRemovedArticles, batchOperation: addedGames.Count > 20);
+                        foreach (var game in addedGames)
```
JosefNemec, line 663: This should be done via BufferedUpdate (on database) to not call game update events for each game individually.

source/Tests/Playnite.Tests/SortableNameConverterTests.cs
```diff
@@ -0,0 +1,89 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Playnite.Tests
+{
+    [TestFixture]
+    public class SortableNameConverterTests
```
JosefNemec, line 11: Needs tests for null, empty and white space only strings.

source/Playnite/ViewModels/MainViewModelBase.cs
Jeshibu: Do you mean the string x = y = z notation, or the variable to save on Resources.GetString calls? If it's the latter I guess that might be premature optimization.

source/Playnite/SortableNameConverter.cs
Jeshibu: I found it useful to have the part that does the validation (which yes, is the majority of the loop content) collapsible in order to just look at the calculation part. If you want, this can be done by nesting it in an if(validate) too.

source/Playnite.DesktopApp/ViewModels/DesktopAppViewModel.cs
Jeshibu: Could you doublecheck this? I can't find a call to DesktopAppViewModel.ImportEmulatedGames from UpdateLibraryData. You might be thinking of MainViewModelBase.ImportEmulatedGames. Maybe this call to SetSortingNames should be in MainViewModelBase though.

source/Playnite.DesktopApp/ViewModels/SettingsViewModel.cs
```diff
@@ -233,6 +234,73 @@ public RelayCommand<object> WindowClosingCommand
...
+                if (res.Result && !res.SelectedString.IsNullOrEmpty())
+                {
+                    if (settings.GameSortingNameRemovedArticles.Any(a => a.Equals(res.SelectedString, StringComparison.InvariantCultureIgnoreCase)))
+                    {
+                        dialogs.ShowErrorMessage(resources.GetString(LOC.ItemAlreadyExists), string.Empty);
+                    }
+                    else
+                    {
+                        SortingNameRemovedArticles.Add(res.SelectedString);
+                        settings.GameSortingNameRemovedArticles = SortingNameRemovedArticles.ToList();
```
JosefNemec, line 254: This should be imo done only once when saving settings, I don't think it's necessary modifying it unless you are saving those changes. And you can use SortingNameRemovedArticles directly in FillSortingNameForAllGames.

source/Playnite.DesktopApp/ViewModels/DesktopAppViewModel.cs
```diff
@@ -572,6 +621,7 @@ public async void ImportInstalledGames(InstalledGamesViewModel model, string pat
                         Logger.Warn("Skipping metadata download for manually added games, some global task is already in progress.");
                     }
                 }
+                await SetSortingNames(addedGames);
```
JosefNemec, line 624: Code styling, needs empty line after closing `}`.

JosefNemec (commented):
Sorry it's taking so long to review these.

PR #2664, opened 2021-11-22
Let the embedded Browser Javascript to send messages to Playnite

nosyn00b:
I have verified that:

The Fanatical plugin already uses this feature but has a fallback that lets it work anyway even if in a less optimized way. I think this can be useful for anyone trying to have interaction with the integrated CefSharp.


source/Playnite/WebView/OffscreenWebView.cs
```diff
@@ -66,6 +69,13 @@ private void Browser_LoadingStateChanged(object sender, LoadingStateChangedEvent
             NavigationChanged?.Invoke(this, new EventArgs());
         }
 
+        private void Browser_JavascriptMessageReceived(object sender, JavascriptMessageReceivedEventArgs e)
+        {
+
+            JavascriptMessageReceived?.Invoke(this, new WebViewJavascriptMessageReceivedEventArgs { message = e.Message.ToString() });
+            NavigationChanged?.Invoke(this, new EventArgs());
```
JosefNemec, line 76: Why is NavigationChanged being called here? This should only JavascriptMessageReceived.

source/PlayniteSDK/Events/WebViewEvents.cs
```diff
@@ -16,4 +16,17 @@ public class WebViewLoadingChangedEventArgs
...
+    /// <summary>
+    /// Represents message from javascript.
+    /// </summary>
+    public class WebViewJavascriptMessageReceivedEventArgs
+    {
+        /// <summary>
+        /// Gets or sets message from javascript
+        /// </summary>
+        /// 
+        public string message { get; set; }
```
JosefNemec, line 29: Wrong capitalization.

JosefNemec (commented):
What's practical use for this? What do you need it for when it comes to Fanatical library?

source/Playnite/WebView/OffscreenWebView.cs
nosyn00b: This function shoud be a copy and paste of the WebView version, You're right.

source/PlayniteSDK/Events/WebViewEvents.cs
nosyn00b: Yes, amended.

nosyn00b:
The use of all this, is to let the embetdded CefSharp browser to send message back proactively if something happens async in the page and not in sync as for example with Javascript Execution Called from Playnite, I use this feature for the Fanatical plugin in this way: when the users logs in, the page changes his contents in an async way (no navigation and no page reloads because Fanatical is react based). So I monitor with  JS the page DOM for successful login and send a message back to playnite asking to close the browser window when it really happens. The only other way I found to say to Playnite that someting changed and the user has logged, so browser windows has to be closed,  is to force a navigation (that causes another interceptable event call on Playnite side). This last way is the fallback I use to support the login function without this changes, but it is not a clean way to make things happen, and moreover the user sees the browser navigate to another page without any real reason after login. In general I think the feature to let Async functions to call Playnyte may be useful also fot other (future) plugins, so I submitted the changes here.

JosefNemec:
You probably changed line endings or encoding because diff is completely borked and I can't review this anymore.

nosyn00b:
Ok, Iìll review it submitting a different Pull request, Thx

Issue #2715, opened 2021-12-18
Add an option to close emulated games via key combination


jmvallejo:
### Feature description

I'm thinking on something similar to what retropie does, where you load games using a core and can press a button combination to exit them (Usually "start + select"). I initially thought an overlay would be best for this but that seems way more complicated.

This would be an awesome feature for those of us who want to play with a controller, plugged in to a TV. PC games usually have the option to exit the game whereas emulated games don't.

I've been looking at the code and I think I already have some ideas. If you think this is worth it and would approve if well implemented I'd like to contribute. Please let me know, thanks!

jlafitte:
I would be interested in this too.  I have one set up for my xbox elite 2 in the "Xbox Accessories" app, but nut I also sometimes use regular xbox controllers, Dualsense controller, or Switch controller depending what emulator I'm using.  Some of the emulators have an option for it, but not all of them.  Would be nice to have it built into playnite or maybe a way to add a plugin for it.

jmvallejo:
I submitted a PR a while ago but they were not a fan of it, it required a lot of changes that I wasn't willing to make at that time. I ended up not fully using Playnite because I needed this feature and didn't want to maintain a fork.

JosefNemec:
There are plugins that can provide this functionality currently:

https://playnite.link/addons.html#InputLayer

https://playnite.link/addons.html#PlayniteGameOverlay_fc75626e-ec69-4287-972a-b86298555ebb

jlafitte:
> There are plugins that can provide this functionality currently:

I never found those when I searched before, I'll give them a try.  Thanks.

Update: InputLayer seems to do exactly what I needed without any bloat!

Issue #684, opened 2018-06-21
DirectInput support in Fullscreen Mode


ClearStrelok:
This would also mean Switch Pro Controller support since they both use DirectInput.

DoganM95:
Any news on this? I am unable to navigate in fullscreen mode with a dualshock 4, which is frustrating..

JosefNemec:
No news, you have to use some DirectInput to XInput wrapper until this is implemented.

Ansem93:
Any hope to have this implemented? I have a ps4 controller and wanted to use Playnite as a launcher :(

firinael:
checking in to say this is still sorely needed.

jh0nny2k:
Just bumping this one up. I use DS4 when playing fighters. Thank you

bukkaroo:
Adding support for this feature - DS4Win is ok, but leaves you with XBox prompts in games, when more and more support native PS prompts.

BoxBoy7999:
bumping this cause ps5 controller no work.

ChillyEmber:
bump, Switch controller doesn't work.

JollyJive:
PS4 controllers don't work for me, even if i map it to a generic input controller with Steam Big Picture's Desktop Configuration.
I don't want to download extra firmware for this, almost every single game these days support PS4 controllers natively and perfectly, I don't want to potentially mess with everything that already works with extra firmware.

Why has this not been implemented yet? After 4 years of this issue no developer has even commented on it
Is it just impossible to implement?

cnapolit:
For the sake of preserving relevant thoughts from discord, I'm posting this here as well.
It could be worth passing off the controller state (both DirectInput & XInput) to plugins so they don't have to poll themselves.

S34NN:
> Adding support for this feature - DS4Win is ok, but leaves you with XBox prompts in games, when more and more support native PS prompts.

I recently came up with my own solution to this by using Special Actions within DS4Windows to switch between profiles on the fly with button combinations. I have mine set up with the default Xinput profile (emulated Xbox 360) and profile emulating a Dualshock 4. When using fullscreen Playnite I can switch to the Xinput profile to navigate it and launch a game, and then quickly switch to the Dualshock 4 profile before the game loads. 

For example I have my profile switching set to PS Button + R1 + X (Cross) to switch to Default Xinput. And PS Button + R1 + Square to switch to the Dualshock 4 profile. 

[...]

PR #2716, opened 2021-12-19
[feat] Feature/allow exiting emulated game from gamepad

jmvallejo:
I'm adding the ability to close games from the gamepad, as requested (by myself) here https://github.com/JosefNemec/Playnite/issues/2715

I'm thinking about where an option for this should go, it can probably be in the emulator profile, but want to get the behavior validated first.

Changes:
- Add process attribute to GenericGameController
- Add StartGamepadPolling method
- Add jmvallejo to contributors.txt

I have verified that:

### DIFF
--- a/source/Playnite.DesktopApp/Resources/contributors.txt
+++ b/source/Playnite.DesktopApp/Resources/contributors.txt
@@ -44,3 +44,4 @@ greggameplayer
 felixkmh
 Curt Grimes
 Jeshibu
+jmvallejo
--- a/source/Playnite/Controllers/GenericGameController.cs
+++ b/source/Playnite/Controllers/GenericGameController.cs
@@ -3,8 +3,6 @@ using Playnite.Database;
 using Playnite.Emulators;
 using Playnite.Scripting.PowerShell;
 using Playnite.SDK;
-using Playnite.SDK.Events;
-using Playnite.SDK.Exceptions;
 using Playnite.SDK.Models;
 using Playnite.SDK.Plugins;
 using System;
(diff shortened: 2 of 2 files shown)


jmvallejo:
Can anyone please take a look at this one?

darklinkpower:
Crow plans to take a look at open PRs during the holidays

From discord:

image

JosefNemec:
I have multiple issues with this:
- it should not be single button hardcoded thing. There must be an option to enable/disable this and option to configure input binding.
  - especially just single button binding is imo very dangerous and could be easily executed by a mistake.
  - actions like this should be generally invoked on button release, not press

- killing process:
  - killing is not good since it can lead to data corruption. You should be sending termination signal and killing process only if that fails to exit the app.
  - killing may fail if started process is elevated.
  - you are not checking if the original process exists. The original process can in some cases start a child, exit and no longer exists when you go to kill it (for example when an emulator updates itself).
  - you are still processing input even after you kill the process

I also think this should be generally done after #684 is implemented, because it will implement interface for general input binding and better handling of various parts of the app using controller input.

jmvallejo:
> I have multiple issues with this:
> 
> * it should not be single button hardcoded thing. There must be an option to enable/disable this and option to configure input binding.
>   
>   * especially just single button binding is imo very dangerous and could be easily executed by a mistake.
>   * actions like this should be generally invoked on button release, not press
> * killing process:
>   
>   * killing is not good since it can lead to data corruption. You should be sending termination signal and killing process only if that fails to exit the app.
>   * killing may fail if started process is elevated.
>   * you are not checking if the original process exists. The original process can in some cases start a child, exit and no longer exists when you go to kill it (for example when an emulator updates itself).
>   * you are still processing input even after you kill the process
> 
> I also think this should be generally done after #684 is implemented, because it will implement interface for general input binding and better handling of various parts of the app using controller input.

Awesome, thanks a lot for the detailed review, I'll try to make this one work according to your feedback after #684.

Cheers!

PR #2793, opened 2022-02-04
Parse doubles according to user culture

### LINKED ISSUE #2789: Grid View "Scroll amount modifier" doesn't accept decimal numbers
### Bug Description

I saw the bug report on Discord and tried it on my Playnite instance, because I suspected it might be related to regional formatting of decimal numbers.

On the one hand, I can reproduce the bug and am now unable to set the scroll amount modifier back to 1.5.
But on the other, it does not appear to be a formatting issue, because neither `1.5` nor `1,5` work.

I tried it with both a custom and the default theme. Same results.
Disabling all of my extensions similarly didn't affect the issue.

### To Reproduce

_No response_

### Diagnostics ID

13e11199-72e8-41c3-a513-c6b0d228258a

### Screenshots

https://user-images.githubusercontent.com/9725089/152147958-9c1d6c6a-da19-4cb7-8747-c81dc36c2a46.mp4

UrbanCMC:
This fixes #2789 by converting both possible separator characters to the one used by the user's current culture.

This is a somewhat naive solution, because it basically only extends the previous behavior to work for all cultures, which means it won't work for numbers that also contain a thousands separator.

This is probably not going to be an issue, but let me know if you would prefer a more accurate solution, which will require the user to use the correct decimal separator.

I have verified that:

### DIFF
--- a/source/Playnite.DesktopApp/Controls/NumericDoubleBox.cs
+++ b/source/Playnite.DesktopApp/Controls/NumericDoubleBox.cs
@@ -1,18 +1,10 @@
 ﻿using Playnite.Common;
 using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
+using System.Globalization;
 using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Data;
-using System.Windows.Documents;
 using System.Windows.Input;
-using System.Windows.Media;
-using System.Windows.Media.Imaging;
-using System.Windows.Navigation;
-using System.Windows.Shapes;
 
 namespace Playnite.DesktopApp.Controls
 {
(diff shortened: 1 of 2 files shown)


source/Playnite.DesktopApp/Controls/NumericDoubleBox.cs
```diff
@@ -1,18 +1,10 @@
...
-using System.Threading.Tasks;
+using System.Globalization;
 using System.Windows;
 using System.Windows.Controls;
 using System.Windows.Data;
-using System.Windows.Documents;
 using System.Windows.Input;
-using System.Windows.Media;
-using System.Windows.Media.Imaging;
-using System.Windows.Navigation;
```
JosefNemec, line 14: Please don't remove usings unnecessarily unless it's a required change related to the fix.

UrbanCMC: No problem. My VS is setup to automatically clean them, but I'll keep it in mind for the future.

Issue #1141, opened 2019-05-08
Port to .NET Core


JosefNemec:
Probably not 3.0 that should release later this year (mainly because CefSharp support is till in the air), but .NET 5 (targeted for late 2020) seems like a logical choice to me.

rogersachan:
https://github.com/cefsharp/CefSharp/issues/2796 Seems to be getting closer.

rogersachan:
> `hez2010` It's already possible to use the current packages with .Net Core 3.0. All the details are listed above. Credit goes to `kpreisser`
> This issue only remains open as there are still some improvements to make integration easier.

Are we waiting for the integration to improve now?

JosefNemec:
I'm not going to port to Core 3.0. I don't think it's a good idea to use such fresh release. Current plan is to port to .NET 5, which should release next year (and be more battle tested).

nkarl7:
How is this coming along? I'm willing to help, if needed

JosefNemec:
I have plans to start looking into it more after .NET 6 is out. Don't need help with it really, since it's a critical change and I want to have full control over it.

shakeyourbunny:
Any news on this, as .NET 6 has been released. 

My guess this is some corner stone for cross platform compatibility.

JosefNemec:
No news, it won't happen sooner than next year (most likely with 6.1). It's not corner stone for cross platform compatibility, see #59

Lancelotbronner:
Was wondering how things were going with regards to this issue and if I could be of any help?

JosefNemec:
Planned for Playnite 11. I'm currently not looking for any help with this, but thanks for offering.

nathenxbrewer:
How is this coming along? I am a .NET developer and can definitely help get things ported to .NET 6 / 7. Once ported, the UI stack should be able to be migrated to .NET MAUI to support MacOS and Linux (Linux is community supported for .NET MAUI.). I'd love to help if you'd let me.

JosefNemec:
Porting to .NET 7 will be done in P11. I'm currently not seeking any help with it, but thanks for the offer. We will not be changing UI stack in 11. Also we won't be using MAUI, the plan is to use Avalonia in future.

nyabinary:
> Porting to .NET 7 will be done in P11. I'm currently not seeking any help with it, but thanks for the offer. We will not be changing UI stack in 11. Also we won't be using MAUI, the plan is to use Avalonia in future.

[...]

PR #2798, opened 2022-02-05
.csproj upgrade to SDK-styled projects

erri120:
WIP

Some notes:
- [`Microsoft.NETFramework.ReferenceAssemblies`](https://github.com/microsoft/dotnet/tree/master/releases/reference-assemblies) enables building without requiring admin operations to install pre-requisites such as Visual Studio or .NET Framework targeting packs
- cleaned up the `AssemblyInfo.cs` files but this can be further simplified by using something like `Directory.Build.props`
- `System.Management.Automation` for `Playnite.csproj` was hard to get, the NuGet packages for this are hidden or target newer .NET versions
- The inlined assemblies in the `references` folders should be replaced by NuGet packages
- similar as the last point: the COM Reference for the Windows Script Host Runtime Library `IWshRuntimeLibrary` should be replaced with some NuGet package
- I think `ApplicationManifest` got yeeted: https://github.com/dotnet/sdk/issues/23627 not sure if needed


JosefNemec:
This should be done only during #1141

JosefNemec:
To add to my last comment, I would prefer to do critical changes like this myself, since I'm in the end the only maintainer of this project.

PR #2821, opened 2022-03-01
Playtime sync rework

darklinkpower:
I have verified that:
- [x] These changes work, by building the application and testing them. I can't test at the moment due to the webview component failing so I can't login to test. This should be considered a draft. **Edit:Tested**

image

I've removed the `LOCSettingsForceDownloadPlaynite` and `LOCSettingsForceDownloadPlayniteTooltip` localization string since they are no longer used with this change

image

image


darklinkpower:
Unless changes are needed the PR is complete. Draft status removed.

source/Tests/Playnite.Tests/Database/GameLibraryTests.cs
```diff
@@ -37,20 +37,20 @@ public void PlaytimeImportTest()
             using (var token = new CancellationTokenSource())
             {
                 db.OpenDatabase();
-                db.ImportGames(libPlugin.Object, true, token.Token);
+                db.ImportGames(libPlugin.Object, token.Token, PlaytimeImportMode.Always);
```
JosefNemec, line 40: Please add test for new third state.

Issue #2786, opened 2022-02-01
Improvements to play action selection window


akumenon:
### Feature description

Henlo,

I'd like to request the following improvements to the play action selection window in desktop mode:
1. Navigation via keyboard buttons (arrows to move, enter to select, escape to back out already works).
2. Scroll bar for when the number of play actions is high (e.g. when you use many different mods).

These are already available in full-screen mode, so desktop mode being at parity is desirable in my opinion.

Thank you for your hard work.
Best regards.

P.S. Probably related to #272

V0idhead:
I'd like to implement this.
And also add the Number keys (1-9) to start the corresponding play action as a third point.

JosefNemec:
Done in devel.

PR #2888, opened 2022-04-14
keyboard navigation in ActionSelectionWindow

V0idhead:
Implemented #2786

In all honesty I have no idea why adding a name to the style for ScrollViewer enables the up and down keys, but without a name they would not select the items like the keys are supposed to do by default.

I have verified that:

### DIFF
--- a/source/Playnite.DesktopApp/Windows/EmulatedGameImportWindow.xaml
+++ b/source/Playnite.DesktopApp/Windows/EmulatedGameImportWindow.xaml
@@ -191,6 +191,8 @@
                                           IsChecked="{Binding ImportWithRelativePaths}"/>
                                 <CheckBox Content="{DynamicResource LOCEmuImportScanSubfolders}" Margin="0,10,0,0"
                                           IsChecked="{Binding ScanSubfolders}"/>
+                                <CheckBox Content="{DynamicResource LOCEmuImportScanInsideArchives}" Margin="0,10,0,0"
+                                          IsChecked="{Binding ScanInsideArchives}"/>
                                 <CheckBox Content="{DynamicResource LOCEmuSaveScanConfig}"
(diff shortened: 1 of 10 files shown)


akumenon:
Wow thank you kindly!

JosefNemec:
Thanks for the contribution. Would it be possible to rework this so it follows MVVM design pattern like the rest of Playnite project?

V0idhead:
Do you mean creating the key bindings for the number keys? I tried to stay as close as possible to the existing code but I didn't find a cleaner way to handle this case.
I didn't want that code in the model, because it's only for the desktop version.
And setting the up the Keybindings in xaml is tricky. Creating 9 hardcoded Bindings doesn't seem like a good idea. And dynamically creating them in xaml would require basically duplicating the ItemsControl-"loop" as well as having a count-variable in the model.

But my experience with WPF is limited.
If you have suggestions on how to do this properly I'll be happy to look into them. Alternatively I can remove this from the pull request for now, so the other changes get through and then I'll look into handling the number keys separately more thoroughly.

JosefNemec:
I've pushed 529caa62ee857559b160443447717e560890396c as an example how I would like to have it done.

- Missing name of `ScrollContentPresenter` is actually a bug, good find. Without it WPF won't setup keyboard inputs properly because it doesn't know what element is content host. We do set it in Fullscreen mode's ScrollViewer, but not in Desktop mode. I don't know why it was missing.
- Focused button on getting highlight is another bug, but it should be fixed in base button style, not just on this view.

The only thing that's not in my changes is binding to number keys. I don't think that's possible to do completely via MVVM since `InputBindings` can't be bound. If you really want to add this functionality, it shouldn't reference underlining view model at all and should use just view data. So probably get child buttons from the ItemsControl and set input bindings based on those.

PR #2908, opened 2022-04-29
keyboard navigation in ActionSelectionWindow part 2

V0idhead:
Continuation / correction of https://github.com/JosefNemec/Playnite/pull/2888
I removed the reference to the view model like you suggested. I'm not completely happy with it because now it depends on the xaml-structure (potential future changes to the DataTemplate making the button not be the first element will result in a crash at runtime). But I guess there is currently no way to implement this nicely.

I have verified that:


JosefNemec:
Won't merge anymore due to conflict, but I've implement it myself already in devel so I'm closing this.

PR #2920, opened 2022-05-08
Ignore config, redistributables and game engine executables during executables scan #2919

### LINKED ISSUE #2919: Library scanner does not exclude most common unrelated executables
### Bug Description

When scanning a library, executables such as installers, unity crash handlers, C++ Redistributables, notification helpers, python executables, are not filtered.
List of executables that should be filtered:

- python.exe    #common in python games
- pythonw.exe    #common in python games
- UnityCrashHandler64.exe    #common in unity games
- UnityCrashHandler32.exe    #common in unity games
- vc_redist.x64.exe    #Microsoft redistributable C++
- vc_redist.x86.exe    #Microsoft redistributable C++
- Config.exe    #clearly a config
- zsync.exe    #updater that seems to be in all RPGM games? Maybe wolf, but still very common
- zsyncmake.exe     #updater that seems to be in all RPGM games? Maybe wolf, but still very common
- notification_helper.exe    #common in some nodejs programs

Screenshots not included for privacy purposes, however attention to case and precise spelling was paid when writing these executables.

### To Reproduce

Scan a folder to find the libraries.

### Diagnostics ID

edbeb8f2-95c3-4e1d-8fcc-a76d5e2b1a70

### Screenshots

_No response_

darklinkpower:
I have verified that:

Closes #2919

devenv_eSEXkqTWHq

### DIFF
--- a/source/Playnite/Common/Programs.cs
+++ b/source/Playnite/Common/Programs.cs
@@ -56,11 +56,40 @@ namespace Playnite.Common
             @"unins\d+"
         };
 
+        private static readonly string[] configAndRedistsExecutablesMasks = new string[]
+        {
+            "Config",
+            "DXSETUP",
+            @"vc_redist\.x64",
+            @"vc_redist\.x86",
+        };
+
+        private static readonly string[] engineExecutablesMasks = new string[]
+        {
+            "UnityCrashHandler32.exe",
+            "UnityCrashHandler64.exe",
+            "notification_helper.exe",
+            "python.exe",
+            "pythonw.exe",
+            "zsync.exe",
+            "zsyncmake.exe"
+        };
+
         private static ILogger logger = LogManager.GetLogger();
 
         public static bool IsFileUninstaller(string path)
         {
             return uninstallerMasks.Any(a => Regex.IsMatch(path, a, RegexOptions.IgnoreCase));
+        }
+
+        public static bool IsFileConfigOrRedistributable(string path)
+        {
+            return configAndRedistsExecutablesMasks.Any(a => Regex.IsMatch(path, a, RegexOptions.IgnoreCase));
+        }
+
+        public static bool IsFileEngineExecutable(string path)
+        {
+            return engineExecutablesMasks.Any(a => a.Equals(path, StringComparison.OrdinalIgnoreCase));
         }
 
         public static void CreateUrlShortcut(string url, string iconPath, string shortcutPath)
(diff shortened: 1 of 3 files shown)


oromis995:
Wow I am impressed! Nice Job!

source/Playnite/Common/Programs.cs
```diff
@@ -56,13 +56,42 @@ public partial class Programs
             @"unins\d+"
         };
 
+        private static readonly string[] configAndRedistsExecutablesMasks = new string[]
```
JosefNemec, line 59: I don't see a good reason to have two separate lists. You can have both set up to be regex matches.

darklinkpower: I thought it would make it easier to maintain, in case more executables are added in the future and that it would also make it clearer to understand what each executable is. It's also the reason why I didn't put the new masks in the existing list for uninstallers. I also think that using a regex for the specific engine executables is not needed since we can know the exact name of the executables. But let me know if you still think that it should use a new single list to change the PR

source/Playnite/Common/Programs2.cs
```diff
@@ -211,21 +201,8 @@ public static async Task<List<Program>> GetShortcutProgramsFromFolder(string pat
...
-                    // Ignore config and redistributables executables
-                    if (IsFileConfigOrRedistributable(targetFileName))
-                    {
-                        continue;
-                    }
-
-                    // Ignore crash handlers and game engine executables
-                    if (IsFileEngineExecutable(targetFileName))
+                    // Ignore uninstallers, config, redistributables and game engine executables
+                    if (IsFileUninstaller(Path.GetFileName(target)))
```
JosefNemec: IsFileUninstaller should be renamed to something more general since it's now not just about uninstaller files.

darklinkpower: huh. That's so weird, I already changed that in https://github.com/JosefNemec/Playnite/pull/2920/commits/f7753378c3dbe3cfdf37a432c241b9f7ec4145a0

I don't know why you still can see it but that name is no longer there

image

vesper8:
@JosefNemec @darklinkpower any chance this feature could be enhanced further by having it also filter out based on a file containing further .exes to ignore when scanning? It could be just a text file in the root of Playnite and we could add one item per line and if any scanned exes match those names exactly then they would not appear in the list of exes to choose from when using the 'scan folder' feature. That would be a nice addition IMO. Thanks for considering it.

JosefNemec:
There is a plan to add support for `.playnite` (similar to .gitignore) files that could be used to configure game import by storing that file with specific settings inside a folder. No specifics yet, probably will be done after P11 is done.

Issue #2504, opened 2021-09-27
Storage manager


JosefNemec:
Basically the same thing as Steam's new storage manager.  #167 should be done together with this.

![image](https://user-images.githubusercontent.com/3874087/134854456-1e62b66d-b34d-41d4-8f35-5453a2cb9a1c.png)

erri120:
I've got some requests to update my [Game Mangement](https://github.com/erri120/Playnite.Extensions/tree/master/src/GameManagement) extension which implements this in a very hacky way:

- size calculations happen at [application startup](https://github.com/erri120/Playnite.Extensions/blob/307a78a682880205bc95cb00f9caf880d0ceaa37/src/GameManagement/GameManagementPlugin.cs#L118) (non-blocking) and when a new game is [installed](https://github.com/erri120/Playnite.Extensions/blob/307a78a682880205bc95cb00f9caf880d0ceaa37/src/GameManagement/GameManagementPlugin.cs#L134)/[uninstalled](https://github.com/erri120/Playnite.Extensions/blob/307a78a682880205bc95cb00f9caf880d0ceaa37/src/GameManagement/GameManagementPlugin.cs#L140)
- size calculation is simple [file enumeration and summation of file sizes](https://github.com/erri120/Playnite.Extensions/blob/307a78a682880205bc95cb00f9caf880d0ceaa37/src/GameManagement/StorageInfo.cs#L57) and are [processed in parallel](https://github.com/erri120/Playnite.Extensions/blob/307a78a682880205bc95cb00f9caf880d0ceaa37/src/GameManagement/StorageInfo.cs#L79) using the [Microsoft High Performance Toolkit](https://docs.microsoft.com/en-us/windows/communitytoolkit/high-performance/parallelhelper)
- storage info is [saved as json](https://github.com/erri120/Playnite.Extensions/blob/307a78a682880205bc95cb00f9caf880d0ceaa37/src/GameManagement/StorageInfo.cs#L41) and displayed in a very ugly and underdeveloped [`DataGrid`](https://github.com/erri120/Playnite.Extensions/blob/307a78a682880205bc95cb00f9caf880d0ceaa37/src/GameManagement/StorageStatisticsView.xaml#L21)

[...]

PR #2928, opened 2022-05-16
Draft implementation of #2504

erri120:
Implements #2504


JosefNemec:
I read you comments in #2504 and please consider not continuing with this, because I'm not going to accept solutions that are incomplete (since you mentioned you are not going to do any UI) or done in, as you said, "very hacky way".

erri120:
cool

PR #3009, opened 2022-08-03
Fix not possible to use TextWrap in CheckBox control

darklinkpower:
I have verified that:

Currently it's not possible to use Textwrap in checkboxes, even when setting the control to do so like this.

```xml
<CheckBox IsChecked="{Binding Settings.Option2}">
    <TextBlock>
        <AccessText TextWrapping="Wrap" 
                    Text="This is a long piece of text attached to a checkbox This is a long piece of text attached to a checkbox This is a long piece of text attached to a checkbox This is a long piece of text attached to a checkbox"/>
    </TextBlock>
</CheckBox>
```

This is caused by the control using a StackPanel. Changing to a DockPanel fixes it.

Before:
Playnite DesktopApp_C2SqSTopIl

After:
RXSppTtNE3

### DIFF
--- a/source/Playnite.DesktopApp/Themes/Desktop/Default/DefaultControls/CheckBox.xaml
+++ b/source/Playnite.DesktopApp/Themes/Desktop/Default/DefaultControls/CheckBox.xaml
@@ -17,8 +17,9 @@
                 <ControlTemplate TargetType="{x:Type CheckBox}">
                     <Border Background="{TemplateBinding Background}"
                             Padding="{TemplateBinding Padding}">
-                        <StackPanel Orientation="Horizontal">
+                        <DockPanel>
                             <Border x:Name="BulletBorder" CornerRadius="{DynamicResource ControlCornerRadius}"
+                                    VerticalAlignment="Center"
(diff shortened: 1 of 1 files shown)


source/Playnite.DesktopApp/Themes/Desktop/Default/DefaultControls/CheckBox.xaml
```diff
@@ -17,8 +17,9 @@
                 <ControlTemplate TargetType="{x:Type CheckBox}">
                     <Border Background="{TemplateBinding Background}"
                             Padding="{TemplateBinding Padding}">
-                        <StackPanel Orientation="Horizontal">
+                        <DockPanel>
                             <Border x:Name="BulletBorder" CornerRadius="{DynamicResource ControlCornerRadius}"
```
JosefNemec, line 21: Children elements should have DockPanel.Dock set to specific values so it's clear how they are docked.

darklinkpower: Done

Issue #1150, opened 2019-05-17
Count when selecting multiple games


unrealed:
It would be nice to have a count of the number of games selected when selecting multiple games. This isn't currently happening in any of the 3 view modes (list, covers, and details). This count could be at the top or at the bottom of the screen.

MrPetovan:
I'm bumping this, I'm curious about the number of games left in my backlog, I have about a thousand games total and about a third played, so I can't really count the resulting list by hand once I have filtered them out.

redactedscribe:
> There is a workaround for this using the Game Media Tools extension by darklinkpower

Would be nice to see this counter/total natively in the client. For example: at the top of Playnite's UI it reads "Filter Active". A " (<shown_games>/<total_games>)" could be appended to this string.

I'm not too interested in the selected number of games (i.e. via <kbd>Shift</kbd> or <kbd>Ctrl</kbd>), but maybe this value could be shown when choosing "Edit..." in the "Edit Game Details" title bar: "Edit Game Details (<selected_games_count> games)".

e5pe:
You can do this too: https://www.reddit.com/r/playnite/comments/aubud9/total_games/ But you have to group the games, for example by platform.

KazuShuSora:
I know that there are workarounds, but I'd still rather have the number of items matching the filter displayed below the drop down menu, for example.
I just want to be able to instantly see how many games match the filter, without having to group the games per platform. Plus, I'd need to manually edit all my DRM-free Humble games for the workaround to work, properly, so that's another reason that'd be annoying.

So if someone could make it possible, that the count of items matching the filter is always displayed at a fixed position, that'd be amazing.

Thank you in advance, if anyone ends up implementing this!

dreampillet:
In case it's not decided yet; in my opinion, the little bar at the bottom of the grid view where the zoom slider is would be the perfect spot to put it, like a status bar.
If no filters are active, it'd display the library total, if filtered, the filtered total and if selected, the selected total.
![image](https://github.com/user-attachments/assets/cb2a13d0-3288-4e3f-b343-5a79965ce3f7)

albertoloscerritos:
+1

PR #3021, opened 2022-08-17
Show collection view count in filter panel

darklinkpower:
I have verified that:
- [x ] I added myself into [contributors file](https://github.com/JosefNemec/Playnite/blob/devel/source/Playnite.DesktopApp/Resources/contributors.txt) if I want to receive contribution credit in the application itself.

Draft just as an idea for #1150

In my opinion the display count show go in the filters panel:
- Can be seen in all views
- Non intrusive
- Makes sense because everything in the filter panel has direct control in the number of games that will be displayed

Current issue is that it displays the number of games including duplicates. Ideally it should only show the number of distinct games.

### Screenshots
image


JosefNemec:

PR #3023, opened 2022-08-18
Fix typos

LemmusLemmus:
This pull request should fix some typos I have noticed while reading the documentation. It is worth mentioning that there are also typos in the names of `ICompletionStatusSettignsApi` in `IPlayniteSettingsAPI` and `MainBackgourndBrush` in `Constants.xaml`, but these have not been changed as to not break things.

**Note:**
I unfortunately only used the GitHub online file editor to make the changes, as such:
- Some files have newlines appended to the end of them by the GitHub editor, it should not matter however.
- An unnecessary amount of commits were made (one for each file). I strongly recommend a squash merge.

I am terribly sorry for any inconvenience caused by this.

**I have verified that:**
- [➖] These changes work, by building the application and testing them. - Nothing has been verified. The only non-documentation change is the renaming of the `closingHanled` variables.
- [✔️] Pull request is targeting `devel` branch.
- [✔️] I added myself into [contributors file](https://github.com/JosefNemec/Playnite/blob/devel/source/Playnite.DesktopApp/Resources/contributors.txt) if I want to receive contribution credit in the application itself. - I feel bad about adding myself to the contribution file for simply fixing some typos, but it seems to be the precedent.

### DIFF
--- a/doc/tutorials/playnite9migration.md
+++ b/doc/tutorials/playnite9migration.md
@@ -22,7 +22,7 @@ Model changes
 Multiple changes have been made to various data models, properties removed, renamed, added. This primarily affects `Game` and `Emulator` objects.
 
 > [!NOTE]
-> Since PowerShell is not statically typed language there's no easy way how to find all model changes you need to fix. Please take care and make sure that you fix  and test your code properly, especially if you are writing data into game library, since you could write bad data it if you don't adjust your scripts properly.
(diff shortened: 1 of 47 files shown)


JosefNemec:
Please remove changes made in the actual code, they cause unnecessary changes in the tracking history, plus you didn't even verify they compile so it's not safe change.

PR #3032, opened 2022-08-25
Add TopPanel View Random Game Selector button

darklinkpower:
I have verified that:

### Description

Adds an option to add a button to the TopPanel that selects a random game in the view when activated

### Demo

https://user-images.githubusercontent.com/1389286/186561646-0b9262d8-d25e-48ab-a414-7d7cc13791c2.mp4

### DIFF
--- a/source/Playnite.DesktopApp/ViewModels/DesktopAppViewModel.cs
+++ b/source/Playnite.DesktopApp/ViewModels/DesktopAppViewModel.cs
@@ -43,6 +43,7 @@ namespace Playnite.DesktopApp.ViewModels
         protected bool ignoreCloseActions = false;
         protected bool ignoreSelectionChanges = false;
         private readonly SynchronizationContext context;
+        private readonly Random random = new Random();
         private Controls.LibraryStatistics statsView;
         private Controls.Views.Library libraryView;
         private SearchViewModel currentGlobalSearch;
@@ -930,6 +931,7 @@ namespace Playnite.DesktopApp.ViewModels
             var model = new RandomGameSelectViewModel(
                 Database,
                 GamesView,
+                random,
                 new RandomGameSelectWindowFactory(),
                 Resources);
             model.OpenView();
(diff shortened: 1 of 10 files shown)


source/Playnite.DesktopApp/ViewModels/DesktopAppViewModel.cs
```diff
@@ -43,6 +43,7 @@ public partial class DesktopAppViewModel : MainViewModelBase, IDisposable, IMain
         protected bool ignoreCloseActions = false;
         protected bool ignoreSelectionChanges = false;
         private readonly SynchronizationContext context;
+        private readonly Random random = new Random();
```
JosefNemec, line 46: It would probably make sense to use static `GlobalRandom` class here. Which then can be used in RandomGameSelectViewModel as well.

source/Playnite/Localization/LocSource.xaml
```diff
@@ -936,6 +936,7 @@ If you want to install or uninstall additional integrations, use "Add-ons" optio
     <sys:String x:Key="LOCTopPanelMainButtonMove">Move main menu button to the sidebar</sys:String>
     <sys:String x:Key="LOCTopPanelExplorerSwitch">Explorer panel</sys:String>
     <sys:String x:Key="LOCTopPanelSelectRandomGameButton">Random game picker</sys:String>
+    <sys:String x:Key="LOCTopPanelViewSelectRandomGameButton">View random game selector</sys:String>
```
JosefNemec, line 939: Not sure about this description. It in my opinion sounds like it will open random game selector, especially if you see it as a tooltip on top panel button.

source/Playnite.DesktopApp/ViewModels/DesktopAppViewModel.cs
darklinkpower: Done in https://github.com/JosefNemec/Playnite/pull/3032/commits/3e3b39498df2812adb0187895b6972aa935a0ce3

source/Playnite/Localization/LocSource.xaml
darklinkpower: Done in https://github.com/JosefNemec/Playnite/pull/3032/commits/4ea1b34d63b8e95f02245f732806ab4a1861bbf2

PR #3037, opened 2022-09-01
Documentation: gameScripts.md

sukarn-m:
Added more examples, including:

* Stopping an application that is running with elevated privileges;
* Conditional execution based on whether an application is already running; and
* Conditional global script based on whether a game has a certain tag.

This is just a documentation contribution, so I haven't added myself to the contributors list.

I have verified that:

### DIFF
--- a/doc/manual/gameScripts.md
+++ b/doc/manual/gameScripts.md
@@ -123,11 +123,37 @@ Start-Process "c:\somepath\someapp.exe" "-some arguments"
 Stop-Process -Name "someapp"
 ```
 
-* If the application requires elevated rights to start then you need to start Playnite as admin too, otherwise the `Stop-Process` will fail due to insufficient privileges.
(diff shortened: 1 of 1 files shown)


JosefNemec:
There already is a section about how to stop elevated process and it doesn't require launching separate PowerShell instance:
https://playnite.link/docs/master/manual/gameScripts.html#cant-shutdown-process

sukarn-m:
> There already is a section about how to stop elevated process and it doesn't require launching separate PowerShell instance: https://playnite.link/docs/master/manual/gameScripts.html#cant-shutdown-process

That doesn't work for everything. This is the output I get when trying to stop "LightingService", which is needed for Asus AURA (RGB Lighting), but it triggers Easy Anti-Cheat in some games.

```
> (Get-WmiObject -Class Win32_Process -Filter "name = 'LightingService.exe'").Terminate()

__GENUS          : 2
__CLASS          : __PARAMETERS
__SUPERCLASS     :
__DYNASTY        : __PARAMETERS
__RELPATH        :
__PROPERTY_COUNT : 1
__DERIVATION     : {}
__SERVER         :
__NAMESPACE      :
__PATH           :
ReturnValue      : 2
PSComputerName   :
```

The process remains running after getting the above output. The method I've added in the documentation asks for admin privileges, but it stops the process.

PR #3100, opened 2022-09-25
Scan install size when updating emulation libraries

darklinkpower:
I have verified that:
- [ ] Pull request is targeting `devel` branch. (Targeting `master`)

Scans size of game of imported by scanners during library update

image

Note: Doesn't include scan for games on MainMenu -> Added. That's for another PR...

### DIFF
--- a/source/Playnite/Localization/LocSource.xaml
+++ b/source/Playnite/Localization/LocSource.xaml
@@ -222,6 +222,7 @@ Selecting "add to exclusion list" option will prevent games from being imported 
     <sys:String x:Key="LOCProgressImportinEmulatedGames">Importing emulated games from {0}…</sys:String>
     <sys:String x:Key="LOCProgressLibraryGames">Downloading library updates…</sys:String>
     <sys:String x:Key="LOCProgressScanningGamesInstallSize">Scanning size of games in library…</sys:String>
+    <sys:String x:Key="LOCProgressScanningEmulatedGamesInstallSize">Scanning size of imported emulated games…</sys:String>
     <sys:String x:Key="LOCProgressLibImportFinish">Library update finished</sys:String>
     <sys:String x:Key="LOCProgressReleasingResources">Releasing resources…</sys:String>
     <!--Menu-->
--- a/source/Playnite/Localization/LocSource.xaml
+++ b/source/Playnite/Localization/LocSource.xaml
@@ -222,7 +222,7 @@ Selecting "add to exclusion list" option will prevent games from being imported 
     <sys:String x:Key="LOCProgressImportinEmulatedGames">Importing emulated games from {0}…</sys:String>
     <sys:String x:Key="LOCProgressLibraryGames">Downloading library updates…</sys:String>
     <sys:String x:Key="LOCProgressScanningGamesInstallSize">Scanning size of games in library…</sys:String>
-    <sys:String x:Key="LOCProgressScanningEmulatedGamesInstallSize">Scanning size of imported emulated games…</sys:String>
+    <sys:String x:Key="LOCProgressScanningImportedGamesInstallSize">Scanning size of imported games…</sys:String>
     <sys:String x:Key="LOCProgressLibImportFinish">Library update finished</sys:String>
     <sys:String x:Key="LOCProgressReleasingResources">Releasing resources…</sys:String>
     <!--Menu-->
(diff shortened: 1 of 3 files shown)


source/Playnite/Localization/LocSource.xaml
```diff
@@ -222,6 +222,7 @@ Selecting "add to exclusion list" option will prevent games from being imported
     <sys:String x:Key="LOCProgressImportinEmulatedGames">Importing emulated games from {0}…</sys:String>
     <sys:String x:Key="LOCProgressLibraryGames">Downloading library updates…</sys:String>
     <sys:String x:Key="LOCProgressScanningGamesInstallSize">Scanning size of games in library…</sys:String>
+    <sys:String x:Key="LOCProgressScanningEmulatedGamesInstallSize">Scanning size of imported emulated games…</sys:String>
```
JosefNemec, line 225: I would leave out that "emulated" part. That way we can reuse this string for non-emulated games as well.

source/Playnite/ViewModels/MainViewModelBase.cs
```diff
@@ -586,28 +586,28 @@ public async Task UpdateLibrary(LibraryPlugin plugin)
...
             try
             {
                 ProgressActive = true;
                 ProgressValue = 0;
-                ProgressTotal = Database.Games.Count + 1;
+                ProgressTotal = games.Count + 1;
 
                 Logger.Info($"Starting Library Install Size scan");
-                ProgressStatus = Resources.GetString(LOC.ProgressScanningGamesInstallSize);
+                ProgressStatus = Resources.GetString(progressMessageLocKey);
```
JosefNemec, line 605: I'm not sure this is needed to have it configurable, if general string is "Scanning size of imported games…" it should be universal enough for all scenarios.

source/Playnite/ViewModels/MainViewModelBase.cs
```diff
@@ -567,7 +567,7 @@ public async Task UpdateLibrary(bool metaForNewGames, bool updateIntegrations, b
 
                 if (AppSettings.ScanLibInstallSizeOnLibUpdate)
                 {
-                    UpdateLibraryInstallSizes(token);
+                    UpdateGamesInstallSizes(token, Database.Games.ToList(), LOC.ProgressScanningGamesInstallSize);
```
JosefNemec, line 570: I don't think .ToList is necessary here. Change the argument type to IEnumerable.

source/Playnite/ViewModels/MainViewModelBase.cs
darklinkpower: The issue is that "Imported" would only apply to the actions that this PR added support because it scans the size of games that were newly imported only. In the case of library updates, the same string with "imported" would not apply because it not only scans the size of newly imported games but also of games that are already in library.

source/Playnite/Localization/LocSource.xaml
darklinkpower: Done in https://github.com/JosefNemec/Playnite/pull/3100/commits/bfff4cae6d5ef7773c8616f178a170d5022f690f

source/Playnite/ViewModels/MainViewModelBase.cs
darklinkpower: You are right, done in https://github.com/JosefNemec/Playnite/pull/3100/commits/f431896072f45a32c84b07adad01df74d53de42d

source/Playnite/ViewModels/MainViewModelBase.cs
```diff
@@ -586,28 +586,28 @@ public async Task UpdateLibrary(LibraryPlugin plugin)
...
-                    UpdateLibraryInstallSizes(token);
+                    UpdateGamesInstallSizes(token, Database.Games, LOC.ProgressScanningGamesInstallSize);
                 }
 
                 return addedGames;
             }, AppSettings.DownloadMetadataOnImport);
         }
 
-        public void UpdateLibraryInstallSizes(CancellationToken token)
+        public void UpdateGamesInstallSizes(CancellationToken token, IEnumerable<Game> games, string progressMessageLocKey)
```
JosefNemec, line 596: Please add token cancellation check at the beginning of this method since it could be in cancelled state from when cancellation was initiated during game import process.

source/Playnite/ViewModels/MainViewModelBase.cs
darklinkpower: Done in https://github.com/JosefNemec/Playnite/pull/3100/commits/bfff4cae6d5ef7773c8616f178a170d5022f690f

source/Playnite/ViewModels/MainViewModelBase.cs
darklinkpower: Done in https://github.com/JosefNemec/Playnite/pull/3100/commits/d224016ca7745e02b9ecf2fe2e645932deeb4118

PR #3128, opened 2022-10-09
Option that chooses whether installation size is calculated through rom file or installation folder

### LINKED ISSUE #3093: Add option to choose between calculating install size based on rom file or installation folder
### Feature description

When  calculating install size of ps3 games playnite always calculates it based on the EBOOT.BIN file which causes the games to show up as being a couple og megabytes in size. Example, demon's souls is around 8 gigabytes but playnite the EBOOT.BIN file is only 26 megabytes

### Screenshots

_No response_

rsa16:
I have verified that:

Fixes #3093

### DIFF
--- a/source/Playnite.DesktopApp/Resources/contributors.txt
+++ b/source/Playnite.DesktopApp/Resources/contributors.txt
@@ -1,4 +1,5 @@
-﻿PhrozenByte
+﻿rsa16
+PhrozenByte
 inondle
 L1cardo
 ArtexJay
--- a/source/Playnite.DesktopApp/ViewModels/GameEditViewModel.cs
+++ b/source/Playnite.DesktopApp/ViewModels/GameEditViewModel.cs
@@ -237,6 +237,13 @@ namespace Playnite.DesktopApp.ViewModels
         {
             get
             {
+                if (EditingGame.Roms.HasItems())
+                {
+                    romExists = true;
+                } else
+                {
+                    romExists = false;
+                }    
                 return romExists;
             }
             set
--- a/source/Playnite.DesktopApp/ViewModels/GameEditViewModel.cs
+++ b/source/Playnite.DesktopApp/ViewModels/GameEditViewModel.cs
@@ -232,6 +232,20 @@ namespace Playnite.DesktopApp.ViewModels
             }
         }
 
+        private bool romExists = false;
+        public bool RomExists
+        {
+            get
+            {
+                return romExists;
+            }
+            set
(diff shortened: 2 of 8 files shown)


rsa16:
Also as a side note, I suggest adding some kind of documentation about Playnite itself due to its complicated project layout. I'm not really that good at C# though so I wouldn't know if this layout is standard, it took me a while to understand it and I still don't get parts of it. Maybe you could add descriptions of what goes inside folders and which files are important?

darklinkpower:
We discussed this briefly in the Discord a week or two ago and it could be concluded that a solution that covers all the cases needs to be more extensive. For example, emulators like RPCS3 can install files from the game files, which would be needed to also be included in the calculation. Some emulators require to scan its installation directory. There may be cases where someone would want to scan both roms and installation directory.

Considering that, this is what I've been thinking how the implementation should be to cover these cases. It's possible that additional files should be added too:

image

---

Crow's idea is for emulator definitions to provide how the scan should be done.

image

I think this would be better for the end user but some issues I can see with this approach is that this would need a lot of additional maintaining for each emulator, investigate how they work and implement their specific solution and for some cases, the approach for size scan can be different between formats even in the same emulator. For this reason personally I would go with the implementation I suggested but in any case, this is something that needs to be discussed to implement the best possible solution.

JosefNemec:
@darklinkpower I don't know, but what you propose seems to me like an overkill solution. If you have a case where you need to specify so many extra options just to calculate the size, it might as well be easier to just enter the size manually. I don't know if in the same case you would have a scenario where the size would regularly change, that's imo unlikely.

One thing that I personally don't like much is that install size calculation feature is starting to get quite complex from maintenance point of view, especially if we add all these options, which is imo not worth it compared to the value it brings. A lot of these edge cases can be solved by setting the size manually.

As far as emulated games specifically, I still stand by my original opinion from Discord that info how to calculate the size (or the size itself) should be provided by emulator support.

rsa16:
@darklinkpower I agree with @JosefNemec, your solution is somewhat overkill for something this simple. Imo, the user shouldn't have to go through every game he/she has to simply customize how the installation size is calculated. A simple option to switch between, as this pull request implements, would be the easiest option to maintain and imo the best option for the user.

Having the emulator provide install size is kind of unrealistic though, since each emulator would then have its own different way to find the installation size and we would end up needing to have a database to track each different method.

All of this is opinion though, don't get me wrong.

JosefNemec:

Issue #3107, opened 2022-09-27
Game Install Size Calculation Failing On Long File Paths


Hassiopeia:
### Bug Description

The new file size calculation introduced in Playnite 10 is failing for long file paths. This produces an error and stops the processing of my game libraries. 

The attached error is even truncating the filepath; there's more to the path than it's even showing. 

### To Reproduce

Create a long file path and attempt to scan it with the install size calculator functions of Playnite.

### Diagnostics ID

1010fd53-74ce-4465-9f95-0ed9a925f087

### Screenshots

![image](https://user-images.githubusercontent.com/46950645/192621735-b5012c2b-0333-4060-887a-65a419a660a2.png)

ppvnf:
Updating to 10.3 should solve this

JosefNemec:
They are already on 10.3

ppvnf:
> They are already on 10.3

My bad. I was searching the issue list and thought this one was the same as in https://github.com/JosefNemec/Playnite/issues/3067, but looking closer the error message is slightly different

JosefNemec:
This functionality has nothing to do with any addon. Do you actually get that error dialog during library update? That should definitely not happen, failures during library update should be silent.

You can disable size calculations during library update in general settings.

Hassiopeia:
Thanks, I didn't know it was able to be turned off, I'll do that. I realize it's not an addon in Plyanite 10, but I thought darklinkpower's size calculation addon from Playnite 9 had been integrated into 10. I'm sorry if I misunderstood that. 

I'm getting the error during the size calculation, but I'm noticing that newly purchased games aren't' being imported, and I assumed it was this error that was stopping the import process.

darklinkpower:
Can you extract this test build (Separate from your normal Playnite install), create two games with the same Installation Directory as those games and check if the error happens in Edit -> Installation -> Calculate

https://drive.google.com/file/d/11RabLpmhcl8ZMYf1XC4iHPnaRBtRWWtj/view?usp=sharing 

If there is an error please upload your playnite.log file

complex-hub:
Tried the build on my machine, since I'm having the same error with Need For Speed: Heat, but sadly it still seems to happen. I attached my playnite.log below.

![image](https://user-images.githubusercontent.com/53913349/196050355-9da31054-d088-4c50-a3b8-caa4296fe96e.png)

[playnite.log](https://github.com/JosefNemec/Playnite/files/9795332/playnite.log)

Issue #3135, opened 2022-10-14
Consistent "Install size scan error" messages


Mellified:
### Bug Description

After updating the game libraries, Playnite 10 shows a notification of an install size error. It's not clear why, as Explorer can navigate to the path successfully. Best guess it is related to the path length, which is over 256 characters. Since Windows 10 the path limit can be disabled via the registry, which I have done.

### To Reproduce

Install a game with a very long path length (?)

### Diagnostics ID

0

### Screenshots

![image](https://user-images.githubusercontent.com/81110968/195910809-ad89a445-a157-4839-a58d-291d6255156d.png)

JosefNemec:
Attach diagnostics ID that was requested in issue template.

Mellified:
Unfortunately Playnite is consistently failing to upload the diag package, which is why the ID is not included.

Mellified:
Diag package and log attached.
[diag.zip](https://github.com/JosefNemec/Playnite/files/9789060/diag.zip)
[playnite.log](https://github.com/JosefNemec/Playnite/files/9789063/playnite.log)

PR #3137, opened 2022-10-18
Scan size symlink fixes and longs Path fix

darklinkpower:
I have verified that:
- [ ] Pull request is targeting `devel` branch. (Targeting `master` as it contains the latest changes)

- Fixes general error in long paths with certain number of characters failing (Fixes https://github.com/JosefNemec/Playnite/issues/3107 and https://github.com/JosefNemec/Playnite/issues/3135)
- Fixes Size scan throwing when it game files contains Directory or Files Symlinks that point to targets that don't exist

### DIFF
--- a/source/Playnite/Common/FileSystem.cs
+++ b/source/Playnite/Common/FileSystem.cs
@@ -346,9 +346,9 @@ namespace Playnite.Common
             return fi.Length;
         }
 
-        public static long GetDirectorySize(string path)
+        public static long GetDirectorySize(string path, bool getSizeOnDisk)
         {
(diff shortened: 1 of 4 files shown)


source/Playnite/Common/FileSystem.cs
```diff
@@ -348,27 +348,36 @@ public static long GetFileSize(FileInfo fi)
 
         public static long GetDirectorySize(string path)
```
JosefNemec, line 349: This should be imo consolidate with GetDirectorySizeOnDisk and just using an argument, like with GetDirectorySize.

PR #3148, opened 2022-10-29
Adding DOSBox profile files for easier use

mingdug:
This is just a tiny change to add DOSBox config and download links for easier use (I love DOSBox and hope I didn't screw it up this time)

I have verified that:


JosefNemec:
1) those arguments don't appear to be correct: https://www.dosbox.com/wiki/Usage
2) I don't think we should be importing exe files but conf files instead
3) ImageExtensions are case insensitive

JosefNemec:
Added via 2ab29a7f30d15dc32901ef541dbee9700b303df8

PR #3199, opened 2022-12-06
Add support for MSX 8-bit computer emulation via libretro

wstephenson:
via fMSX and blueMSX libretro plugins.

I have verified that:


JosefNemec:
Was added with other RA updates via #3289

Issue #1592, opened 2019-12-15
Alternative name for easier searching


CorvoJavidan:
Would be nice to have an alternative name entry option so the user can choose which way to search for.

In the game detail page they can add acronyms such as eso/wow/kotor to search for that instead of elder scrolls online/world of warcraft/knights of the old republic. Or ignoring punctuation by searching for "fear" or "asura vengeance" instead of "F.E.A.R." and "Asura: Vengeance"

JosefNemec:
Acronym search is already supported and support for full alternative names is tracked in #3190, closing.

PR #3202, opened 2022-12-10
Acronyms search support

darklinkpower:
I have verified that:

Closes https://github.com/JosefNemec/Playnite/issues/1592

### Implements:
- Support to search items by their acronyms in Global search and game name filters
- Setting to enable this behavior in the Global Search window. Enabled by default.
- Support to enable this behavior in filters for game name searches. Disabled by default since it can potentially provide unwanted results, specially for already saved filters.

### Screenshots

image

image

image

image

### DIFF
--- a/source/Playnite.DesktopApp/Controls/SettingsSections/Search.xaml
+++ b/source/Playnite.DesktopApp/Controls/SettingsSections/Search.xaml
@@ -40,6 +40,11 @@
                           DockPanel.Dock="Top" Margin="0,15,0,0"
                           IsChecked="{Binding Settings.IncludeCommandsInDefaultSearch}" />
 
+                <CheckBox Content="{DynamicResource SearchBoxSearchWithAcronyms}"
+                          ToolTip="{DynamicResource SearchBoxSearchWithAcronymsTooltip}"
+                          DockPanel.Dock="Top" Margin="0,15,0,0"
+                          IsChecked="{Binding Settings.SearchBoxSearchWithAcronyms}" />
+                
                 <StackPanel Orientation="Horizontal" Margin="0,15,0,0" DockPanel.Dock="Top" >
                     <TextBlock VerticalAlignment="Center" Margin="0,0,10,0"
                                Text="{DynamicResource LOCSearchSettingsSystemWideShortcut}"/>
--- a/source/Playnite.DesktopApp/Controls/SettingsSections/Search.xaml
+++ b/source/Playnite.DesktopApp/Controls/SettingsSections/Search.xaml
@@ -39,11 +39,6 @@
                           ToolTip="{DynamicResource LOCSearchIncludeCommandsInDefaultTooltip}"
                           DockPanel.Dock="Top" Margin="0,15,0,0"
                           IsChecked="{Binding Settings.IncludeCommandsInDefaultSearch}" />
-
-                <CheckBox Content="{DynamicResource SearchBoxSearchWithAcronyms}"
-                          ToolTip="{DynamicResource SearchBoxSearchWithAcronymsTooltip}"
-                          DockPanel.Dock="Top" Margin="0,15,0,0"
-                          IsChecked="{Binding Settings.SearchBoxSearchWithAcronyms}" />
                 
                 <StackPanel Orientation="Horizontal" Margin="0,15,0,0" DockPanel.Dock="Top" >
                     <TextBlock VerticalAlignment="Center" Margin="0,0,10,0"
--- a/source/Playnite/Common/Extensions/StringExtensions.cs
+++ b/source/Playnite/Common/Extensions/StringExtensions.cs
@@ -88,7 +88,8 @@ namespace System
 
         public static bool IsStartOfStringAcronym(this string acronymStart, string input)
         {
-            if (string.IsNullOrEmpty(input) || string.IsNullOrEmpty(acronymStart) || acronymStart.Length > input.Length)
+            if (string.IsNullOrEmpty(input) || string.IsNullOrEmpty(acronymStart)
+                || acronymStart.Length > 2 || acronymStart.Length > input.Length)
             {
                 return false;
             }
--- a/source/Playnite/Common/Extensions/StringExtensions.cs
+++ b/source/Playnite/Common/Extensions/StringExtensions.cs
@@ -86,6 +86,38 @@ namespace System
             }
         }
 
+        public static bool IsStartOfStringAcronym(this string acronymStart, string input)
+        {
+            if (string.IsNullOrEmpty(input) || string.IsNullOrEmpty(acronymStart) || acronymStart.Length > input.Length)
+            {
+                return false;
+            }
+
+            var acronymIndex = 0;
+            for (int i = 0; i < input.Length; i++)
+            {
+                if (char.IsLetterOrDigit(input[i]) && (i == 0 || input[i - 1] == ' '))
+                {
+                    if (char.ToUpperInvariant(input[i]) != char.ToUpperInvariant(acronymStart[acronymIndex]))
+                    {
+                        return false;
+                    }
+                    else
+                    {
+                        acronymIndex++;
+                        // If the acronym index and acronym start length is the same
+                        // it means all the characters have been matched
+                        if (acronymIndex == acronymStart.Length)
+                        {
+                            return true;
+                        }
+                    }
+                }
+            }
+
+            return false;
+        }
+
         private static string RemoveUnlessThatEmptiesTheString(string input, string pattern)
         {
             string output = Regex.Replace(input, pattern, string.Empty);
--- a/source/Playnite/Common/Extensions/StringExtensions.cs
+++ b/source/Playnite/Common/Extensions/StringExtensions.cs
@@ -94,6 +94,14 @@ namespace System
                 return false;
             }
 
+            for (int i = 0; i < acronymStart.Length; i++)
+            {
+                if (!char.IsLetterOrDigit(acronymStart[i]))
+                {
+                    return false;
+                }
+            }
+
             var acronymIndex = 0;
             for (int i = 0; i < input.Length; i++)
             {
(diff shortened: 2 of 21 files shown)


source/Playnite/Common/Extensions/StringExtensions.cs
```diff
@@ -86,6 +86,38 @@ public static string ToTileCase(this string source, CultureInfo culture = null)
             }
         }
 
+        public static bool IsStartOfStringAcronym(this string acronymStart, string input)
+        {
+            if (string.IsNullOrEmpty(input) || string.IsNullOrEmpty(acronymStart) || acronymStart.Length > input.Length)
```
JosefNemec, line 91: This should also exit if acronymStart is only one character long.

source/Playnite/Database/GameDatabase_Filters.cs
```diff
@@ -840,5 +812,32 @@ private bool FilterByStyleOr(Game game, FilterSettings filterSettings)
...
+            {
+                return false;
+            }
+
+            if (filterSettings.NameSearchWithAcronyms &&
+                filterSettings.Name.IsStartOfStringAcronym(game.Name))
+            {
+                return true;
+            }
+            else if (filterSettings.Name.Length >= 2 && filterSettings.Name[0] == '^')
```
JosefNemec, line 828: This check should be done first because of explorer panel filter via name groups.

source/Playnite/Localization/LocSource.xaml
```diff
@@ -1172,6 +1172,8 @@ SHIFT-ENTER: open item menu</sys:String>
     <sys:String x:Key="LOCEmuOverridePlatformTooltip">When set, scanner will assign this platform to all games, overwriting any automatically detected platforms.</sys:String>
     <sys:String x:Key="LOCSearchIncludeCommandsInDefault">Include commands in default search</sys:String>
     <sys:String x:Key="LOCSearchIncludeCommandsInDefaultTooltip">When disabled, commands won't be included in default search until # prefix is used.</sys:String>
+    <sys:String x:Key="SearchBoxSearchWithAcronyms">Search using acronyms</sys:String>
```
JosefNemec, line 1175: In my opinion this should be called something like `Support search using acronyms`. At first glance it looks like the search will be done only using acronyms when enabled.

source/Playnite/ViewModels/SearchViewModel.cs
```diff
@@ -117,7 +117,7 @@ private bool GameFilter(Game game, string searchTerm, GameSearchFilterSettings s
                 return false;
             }
 
-            if (!SearchViewModel.MatchTextFilter(searchTerm, game.Name))
+            if (!SearchViewModel.MatchTextFilter(searchTerm, game.Name, mainModel.AppSettings.SearchBoxSearchWithAcronyms))
```
JosefNemec, line 120: This should be instead passed into the method as an argument, for easier testing in future.

source/Playnite/ViewModels/SearchViewModel.cs
```diff
@@ -655,7 +655,7 @@ private List<SearchItem> FilterSearchResults(List<SearchItem> toFilter, string f
             return results;
         }
 
-        public static bool MatchTextFilter(string filter, string toMatch)
+        public static bool MatchTextFilter(string filter, string toMatch, bool matchTargetAcronymStart = false)
```
JosefNemec, line 658: matchTargetAcronymStart shouldn't have default value so it's clear how the match is actually done.

source/Tests/Playnite.Tests/SearchViewModelTests.cs
```diff
@@ -28,6 +28,16 @@ public void MatchTextFilterTest()
             Assert.IsTrue(SearchViewModel.MatchTextFilter(null, null));
             Assert.IsTrue(SearchViewModel.MatchTextFilter("", null));
             Assert.IsTrue(SearchViewModel.MatchTextFilter(null, ""));
+
+            // Acronym tests
+            Assert.IsTrue(SearchViewModel.MatchTextFilter("gow", "Gears of War", true));
```
JosefNemec, line 33: These tests should include strings with random additional spaces in both arguments (`g OW` and ` god   of war 3 ` for example) and also tests with acronym search disabled.

source/Playnite/Common/Extensions/StringExtensions.cs
```diff
@@ -86,6 +86,38 @@ public static string ToTileCase(this string source, CultureInfo culture = null)
             }
         }
 
+        public static bool IsStartOfStringAcronym(this string acronymStart, string input)
+        {
+            if (string.IsNullOrEmpty(input) || string.IsNullOrEmpty(acronymStart) || acronymStart.Length > input.Length)
+            {
```
JosefNemec, line 92: Also should probably return if acronymStart contains white space.

source/Playnite/Common/Extensions/StringExtensions.cs
darklinkpower: Fixed in https://github.com/JosefNemec/Playnite/pull/3202/commits/f5f00f9913c5c6c57fbc6710f7ddfba519be2dce.

I couldn't find info about it but I think it makes sense to limit it to at least 2 since acronyms are words which requires having multiple characters at least.

source/Playnite/Localization/LocSource.xaml
darklinkpower: Changed

source/Playnite/ViewModels/SearchViewModel.cs
darklinkpower: Done in https://github.com/JosefNemec/Playnite/pull/3202/commits/46edfe15bb973655782d00d3889b2c5442b83c7e

source/Playnite/Common/Extensions/StringExtensions.cs
darklinkpower: Done in https://github.com/JosefNemec/Playnite/pull/3202/commits/ca68783991e9974784db3a8909bbf8a0fc8b7288

source/Playnite/ViewModels/SearchViewModel.cs
darklinkpower: I don't understand since from my understanding that's already how it's being done, can you specify how it should be done if possible?

source/Tests/Playnite.Tests/SearchViewModelTests.cs
```diff
@@ -30,7 +30,6 @@ public void MatchTextFilterTest()
             Assert.IsTrue(SearchViewModel.MatchTextFilter(null, "", false));
 
             // Acronym tests
-            Assert.IsFalse(SearchViewModel.MatchTextFilter("g", "Gears of War", true));
```
darklinkpower: Removed because it will return true because the "g" will get matched by normal string search

source/Playnite/ViewModels/SearchViewModel.cs
JosefNemec: Add new bool argument to GameFilter method and use that argument instead of direct reference to app settings.

source/PlayniteSDK/Database/IGameDatabase.cs
```diff
@@ -202,14 +202,16 @@ public interface IGameDatabase
         /// </summary>
         /// <param name="game"></param>
         /// <param name="filterSettings"></param>
+        /// /// <param name="nameFilterSearchWithAcronyms"></param>
         /// <returns></returns>
-        bool GetGameMatchesFilter(Game game, FilterPresetSettings filterSettings);
+        bool GetGameMatchesFilter(Game game, FilterPresetSettings filterSettings, bool nameFilterSearchWithAcronyms = false);
```
JosefNemec, line 207: This is breaking change. You need to add new separate method overload.

source/Tests/Playnite.Tests/SearchViewModelTests.cs
```diff
@@ -15,30 +15,44 @@ public class SearchViewModelTests
...
+            // Acronym tests
+            Assert.IsTrue(SearchViewModel.MatchTextFilter("gow", "Gears of War", true));
+            Assert.IsTrue(SearchViewModel.MatchTextFilter("goW", " Gears of War", true));
+            Assert.IsTrue(SearchViewModel.MatchTextFilter("GOW", "god of war 3", true));
+            Assert.IsFalse(SearchViewModel.MatchTextFilter(" GOW", "god of war 3", true));
+            Assert.IsFalse(SearchViewModel.MatchTextFilter("GOW ", "god of war 3", true));
+            Assert.IsTrue(SearchViewModel.MatchTextFilter("GOW", "god   of war   3", true));
+            Assert.IsFalse(SearchViewModel.MatchTextFilter("g OW", "god of war 3", true));
+            Assert.IsTrue(SearchViewModel.MatchTextFilter("gOw3", "god of war 3", true));
+            Assert.IsFalse(SearchViewModel.MatchTextFilter("gOw3", "god of war 2", true));
```
JosefNemec, line 41: Missing tests with disabled acronym support I mentioned in my original comment.

source/Playnite.DesktopApp/Controls/SettingsSections/Search.xaml
```diff
@@ -40,6 +40,11 @@
                           DockPanel.Dock="Top" Margin="0,15,0,0"
                           IsChecked="{Binding Settings.IncludeCommandsInDefaultSearch}" />
 
+                <CheckBox Content="{DynamicResource SearchBoxSearchWithAcronyms}"
```
JosefNemec, line 43: As we discussed on Discord, this shouldn't be an option and acronym support should be always enabled.

source/Tests/Playnite.Tests/SearchViewModelTests.cs
darklinkpower: I think I did what you indicate, please let me know if it's okay

source/PlayniteSDK/Database/IGameDatabase.cs
darklinkpower: Fixed in https://github.com/JosefNemec/Playnite/pull/3202/commits/17b0389436ad6c5e08e9624bbacf71e8b5e7c611

Issue #3287, opened 2023-02-04
Filter settings don't save when exiting Playnite in certain ways


darklinkpower:
### Bug Description

Filter settings don't save when exiting Playnite in certain ways.

### To Reproduce

Applied filters don't save when exiting Playnite in certain ways. I've found the following:

- Closing Playnite with `--shutdown` argument
- Closing Playnite after extension or Playnite itself crashes
- Might be misremembering but I think it also has happened when Playnite is closed after starting Windows shutdown or restart process

### Diagnostics ID

51a6cae9-d6ac-4fdc-a39d-fa2b0e1dd9cd

JosefNemec:
> Closing Playnite after extension or Playnite itself crashes and shows built-in crash dialog

I don't think it's save saving settings in this scenario.

> Might be misremembering but I think it also has happened when Playnite is closed after starting Windows shutdown or restart process

Settings are saved in this case: https://github.com/JosefNemec/Playnite/blob/master/source/Playnite/App/PlayniteApplication.cs#L456

darklinkpower:
I agree.

I don't know why but I can't reproduce the isue with command line anymore. I've sent a PR to add logging when saving settings for debugging and troubleshooting purposes

https://github.com/JosefNemec/Playnite/pull/3288

chocolatechipcats:
I've had this happen from power outages turning off the computer, but didn't think to report it.

PR #3288, opened 2023-02-04
Add additional logging when saving settings

darklinkpower:
For troubleshooting purposes

I have verified that:


JosefNemec:
Not sure why this would be helpful, especially since we already log failure.

darklinkpower:
I thought it could be useful to troubleshoot issues like https://github.com/JosefNemec/Playnite/issues/3287 as it would indicate when it's saving but yeah maybe not needed

Issue #3248, opened 2023-01-05
Can HDR and Auto HDR switches be added to Playnite?


shinobiz:
### Feature description

As we all know, the HDR switch of Win10 and Win11 is very inconvenient to control in the system settings. Can HDR and Auto HDR switches be added to Playnite? Especially in full screen mode. 

Thanks programmer!

JosefNemec:
Already possible using game scripts:
https://www.reddit.com/r/playnite/comments/zs3m3d/config_for_simple_hdr_switching/
https://www.reddit.com/r/playnite/comments/d5yccj/using_powershell_to_switch_to_hdr_mode_then/

shinobiz:
Thank you for your reply. But I still wish HDR Switch was integrated into Playnite, not just via script.

从 Windows 版邮件发送

发件人: Josef Nemec
发送时间: 2023年1月5日 15:34
主题: Re: [JosefNemec/Playnite] Can HDR and Auto HDR switches be added toPlaynite? (Issue #3248)

Already possible using game scripts:
https://www.reddit.com/r/playnite/comments/zs3m3d/config_for_simple_hdr_switching/
https://www.reddit.com/r/playnite/comments/d5yccj/using_powershell_to_switch_to_hdr_mode_then/

lscholte:
@shinobiz Your proposal inspired me to experiment with this change because I have also had a need for this feature since I don't like toggling HDR and using start/stop scripts is more cumbersome compared to a simple checkbox that I can quickly set on each game that needs this.

This is what I created based on your mockup above. I should note however, that I have not put any effort to trying to put this in fullscreen mode, but that's largely because I've had no need for it
<img width="634" alt="image" src="https://user-images.githubusercontent.com/6503344/216218404-41d4522c-0057-47ef-8045-7b8cd047c5c5.png">

Ultimately I did was use this checkbox to create some implicit start/stop scripts that run. The script specifically starts up a C++ program I wrote that uses NVAPI (a library for NVIDIA GPUs) to enable HDR and disable it when the program terminates. It works well for me (it even works correctly if HDR is already enabled in that it won't kill HDR when the game exits), but as it is, I'm not sure this is the type of thing I could put in a PR in its current state. The main problem is my reliance on an NVIDIA library which limits who this is useful for.

shinobiz:
[...]

PR #3290, opened 2023-02-05
Activate HDR When Game Launches

lscholte:
I have verified that:

**Description**
This PR implements a feature request https://github.com/JosefNemec/Playnite/issues/3248. It adds an Enable/Disable HDR button to the context menu for each game in both desktop and fullscreen mode. In addition, HDR can be toggled in the advanced options when you edit the game details. The HDR options in the context menu will only be visible if your primary display supports HDR. In the advanced game options, the HDR option will always be visible, but will be disabled if the primary display does not support HDR.

In order to accomplish this, I have added a C++/CLI project which uses some Windows API calls to toggle HDR status on the primary display. Various places in the Playnite C# code can call this.

This is the context menu for a game when HDR is currently enabled for it
image

This is the context menu for a game when HDR is not currently enabled
image

This is in the advanced game options settings when my primary display supports HDR
image

Context menu for a game when my primary display does not support HDR
image

This is in the advanced game options settings when my primary display does not support HDR
image

This is the bulk edit context menu
image

The bulk edit advanced game options
image

Context menu for a game in fullscreen mode when HDR is disabled for the game
image

Context menu for a game in fullscreen mode when HDR is enabled for the game
image

**Outstanding Issues**
I am not aware of anything that is broken. However, I am only capable of testing this on Windows 11.

**Testing Performed**
- I have tested on Windows 11 using both an NVIDIA GPU and Intel Integrated graphics
- When the setting is enabled on a game and HDR is not currently active on the primary display, launching the game will activate HDR if it is not already active and will deactivate HDR when the game exits
- When the setting is enabled on a game and HDR is currently active on the primary display, launching the game will effectively do nothing since HDR is already enabled. HDR will continue to be enabled on the primary display when the game exits
- When my primary monitor does not support HDR at all, launching the game is effectively the same as it would be without this feature
- If I change my primary display to a monitor that does not support HDR, then all context menu options in both fullscreen and desktop mode for HDR disappear
- - In the advanced game options, the HDR setting will be disabled and unchecked and has the correct tooltip
- When the primary monitor does support HDR, then the HDR option in advanced game settings has a tooltip explaining what the setting does
- HDR setting correctly persists when re-launching Playnite
- Setting HDR for one game does not impact any other games
- Bulk edit operations work for HDR


shinobiz:
It looks very nice. I'm really looking forward to experiencing it sooner. Will it be merged into the master version in the future?

JosefNemec:
- I don't think it's a good idea to have this as a game menu option since it's a system wide option. It should be in the main menu.
- I'm really not a fan of bringing in C++/CLI dependency, this is a deal breaker for me. Can't this be done with pinvoke calls only?

JosefNemec:
Please ignore my first comment, I didn't realize it just toggles the HDR flag on games. In which case I'm not sure it even should be an option in game menu. 

Is it something that a user would frequently change instead of setting it once and leaving it? The main issue is that game menu is already pretty crowded, especially with plugins installed, and there's currently no way how to hide unwanted items.

JosefNemec:
To clarify my dislike for C++/CLI in this case:
- we are slowly but surely moving away from Windows only dependencies
- it's a huge dependency for relatively small feature that not that many people will use

If you really need to use it, it might be better to make it as a plugin.

shinobiz:
> To clarify my dislike for C++/CLI in this case:
> 
> * we are slowly but surely moving away from Windows only dependencies
> * it's a huge dependency for relatively small feature that not that many people will use
> 
> If you really need to use it, it might be better to make it as a plugin.

I feel that as HDR monitors and TVs become more affordable, the HDR option will be an important and indispensable feature.

lscholte:
> * I'm really not a fan of bringing in C++/CLI dependency, this is a deal breaker for me. Can't this be done with pinvoke calls only?

It could. My preference was for C++/CLI because I find it to be less cumbersome than doing this in C#, but I can switch that out when I get some time.

lscholte:
> Please ignore my first comment, I didn't realize it just toggles the HDR flag on games. In which case I'm not sure it even should be an option in game menu.
> 
> Is it something that a user would frequently change instead of setting it once and leaving it? The main issue is that game menu is already pretty crowded, especially with plugins installed, and there's currently no way how to hide unwanted items.

It is true that this would be a flag that users are most likely setting once on individual games and then probably never touching again for those games. So yeah I see your point about it maybe being unnecessary in the game menu, although since I added it, I did find it quite convenient. But is that convenience worth it? 

Also, one of the requests was for this to work in full-screen mode too (CC @shinobiz ). In that case, is there an alternative place you think would make sense to place this setting that would allow fullscreen users to toggle it without switching to desktop mode? For me personally at least, it's not really a requirement for me personally for these options to exist in a game context menu.

lscholte:
@JosefNemec If I did the following, do you expect this is something that I could merge? Or do you have deeper concerns about this feature even existing at all?
1) Replace C++/CLI code with C# P/Invoke code
2) Got rid of the context menu options but kept the option in the advanced game dialog

shinobiz:
> To clarify my dislike for C++/CLI in this case:
> 
> * we are slowly but surely moving away from Windows only dependencies
> * it's a huge dependency for relatively small feature that not that many people will use
> 
> If you really need to use it, it might be better to make it as a plugin.

Now maybe you are right, I believe there will be more and more demand for this feature in the next 3~5 years. I think the existence of this feature is also a major advantage compared to similar software. Maybe launch a community vote now to see how many users want this feature.

Also, I don't think it's a good idea to have the HDR option exist as a plugin, since it acts as a marker or label in addition to HDR start and stop. Like tagging game categories, platforms, favorites or hidden information. The user only needs to mark once, and the program can perform the HDR function according to the mark, without the user needing to frequently switch the HDR function in the Windows system settings. A small function improvement greatly improves the user experience. Of course, sometimes, users also want to temporarily turn on or off HDR for a certain game. I think it is very necessary for it to exist in the menu, regardless of desktop mode or full-screen mode. If it only exists as a plugin, when the user accidentally deletes the plugin, these flags will be cleared, and the user is wasting time and energy for it...

shinobiz:
> @JosefNemec If I did the following, do you expect this is something that I could merge? Or do you have deeper concerns about this feature even existing at all?
> 
> 1. Replace C++/CLI code with C# P/Invoke code
> 2. Get rid of the context menu options but keep the option in the advanced game dialog
> 
> EDIT: Also couple things on point number 2. First is that I should re-iterate that these context menu options are only going to be visible for people who have an HDR-capable display as their primary display (i.e those people who would probably find this feature useful), and with Windows 11 there is an auto HDR feature which can add HDR effects even to older games that don't officially support it. So in that sense, it might actually be something people want to experiment with for certain games where they may toggle on and off several times. Lastly -- and this is just an idea -- but we could consider a settings page for the overall Playnite application that allows configuration of the context menus.

Yes, I am the guy you said, I am trying the HDR effect of PS1 PS2 PSP Wii GC DC and other emulators under DX11/DX12, and there are already considerable effects, I think I need HDR function, and whether it is a desktop Still in full screen mode. I also often need to switch the HDR function for comparison, or temporarily turn off the HDR function when recording game videos (to prevent recording errors under HDR), so I think it is necessary to keep the HDR option in the menu.

lscholte:
@JosefNemec I'm working on replacing C++/CLI with P/Invoke, but it will take me a bit of time (guessing a day or two) to do so because
a) There is a lot of boiler plate that I need to write to get this to actually work correctly and it is kinda tedious to do (I genuinely believe C++/CLI is the more practical approach for what I am trying to do)
b) I will need to test it again
c) I work full time already

Once I have it done, I'll put it up into this PR. However, I would really like you to reconsider your stance on C++/CLI once you see what the P/Invoke version looks like. For the record, I actually don't mind writing it because it is good practice for me; it's just I think the end result is going to be more awkward to maintain because it's not as simple as just importing a couple of functions I need and calling them.

JosefNemec:
As I said, if you really want to use C++/CLI, make it a custom plugin. I really don't want to add the entire C++/CLI tool chain dependency just for this one small feature.

lscholte:
@JosefNemec Ok the C++/CLI has been replaced with C# code now.

Removing the Enable/Disable HDR option from the game context menu in desktop mode has not yet been done, but that's on my TODO list for probably tomorrow night. Unless you are ok with keeping that context menu option since it only appears for those who have an HDR monitor.

darklinkpower:
Hi, sorry for meddling but I was checking the PR and found this summary that seems incorrect

image

Should probably say

> Gets or sets value indicating if HDR should be enabled for the primary display while playing the game.

or similar

JosefNemec:
I have not checked these pinvoke changes in detail yet, but I don't see a good reason why it should be a separate project instead of being part of core Playnite project. Am I missing something?

lscholte:
> I have not checked these pinvoke changes in detail yet, but I don't see a good reason why it should be a separate project instead of being part of core Playnite project. Am I missing something?

I did it that way so I can make use of the `internal` access modifier on the `DisplayConfig` class so that the only possible thing the rest of Playnite code can be calling is the `HdrUtilities`. That said, I could stick everything into one file and use `private` instead and accomplish the same thing.

lscholte:
> Hi, sorry for meddling but I was checking the PR and found this summary that seems incorrect
> 
> image
> 
> Should probably say
> 
> > Gets or sets value indicating if HDR should be enabled for the primary display while playing the game.
> 
> or similar

Meddle all you want! I have fixed this

lscholte:
I have now removed the HDR toggle from the game context menu in desktop mode. It remains in the context menu for fullscreen mode. It also remains in the advanced game options dialog in desktop mode.

lscholte:
> I have not checked these pinvoke changes in detail yet, but I don't see a good reason why it should be a separate project instead of being part of core Playnite project. Am I missing something?

I have now moved the P/Invoke stuff and reused a bunch of the things that turned out to already exist.

Let me know if there are other concerns.

Agitatio:
Does it enable HDR globally or only for primary display?

Agitatio:
> Primary display only. In my case, I have 3 monitors, 2 of which support HDR. But when I game, I only actually want my main display to have it enabled because I can use my other displays for non-gaming things and having HDR would detract from that for me.
> 
> That said, it is definitely possible to activate it for all monitors that support it. The only added challenge with that is that when the game exits, it reverts to the HDR state it had before the game was launched, so I'd need to keep track of that state for all monitors.
In my case I game on my monitor and TV depending on the game. In the meantime I've just been using an AHK script that presses Win+Alt+B that is then launched through per-game scripts. Although I would've still preferred and ability to toggle it for specific displays and not just all of them.

lscholte:
> > Primary display only. In my case, I have 3 monitors, 2 of which support HDR. But when I game, I only actually want my main display to have it enabled because I can use my other displays for non-gaming things and having HDR would detract from that for me.
> > That said, it is definitely possible to activate it for all monitors that support it. The only added challenge with that is that when the game exits, it reverts to the HDR state it had before the game was launched, so I'd need to keep track of that state for all monitors.
> 
> In my case I game on my monitor and TV depending on the game. In the meantime I've just been using an AHK script that presses Win+Alt+B that is then launched through per-game scripts. Although I would've still preferred and ability to toggle it for specific displays and not just all of them.

Ok so I can see some possibilities:
1) The HDR setting activates HDR on the primary display. The assumption here is that most people would play games on that display, but it may not be true for everyone. This is the PR that I currently have up.
2) Another possibility is to have the HDR setting activate HDR on all displays that support it. If I went this route, I would want to add an application-wide setting that determines if HDR will be activated on all displays or only the primary display. This is something I'd be open to exploring, but I have no personal use for it at this time.
3) List out the displays that support HDR and create an application-wide setting that allows you to specify which display(s) you want HDR activated on when a game starts. I am also open to exploring this, and honestly I think this is probably the best possible thing to do because it allows everyone to configure this feature to meet their specific needs.

If I were to look into either 2 or 3, I'd be inclined to do it as a follow-up PR simply because I believe what I have right now is already a valuable addition to Playnite and I wouldn't want to get bogged down trying to do too much at once. I'm open to thoughts anyone has on this topic.

ThinkSalat:
Just as an fyi - Special K has this ability baked in - from the hdr menu, you can select to have HDR start up whenever you open the game. Considering there's already a special K extension that works quite well, this could work as a holdover for anyone who'd like this feature but odn't want to wait for the pr to merge.

Also have you considered creating an extension that just adds a tag like `[HDR-Toggle] True`. Open to interpretation of course. I suppose considering you're adding c++ dependencies that it may not work, but an extension would probably be a better bet than a merge into the main repo

darklinkpower:
> Just as an fyi - Special K has this ability baked in [...]

The issue with Special K is in that case you need to rely on external software and also Special K has compatibility issues with games, ranging from failing to launch games, crashes, not being able to be used in games with anti cheat (Or risk being banned), causing graphical issues, etc. For that reason personally I think it's better to have this feature built into Playnite either as a native feature or as an addon.

> suppose considering you're adding c++ dependencies

The PR has been updated since that comment and now it uses C# code only

lscholte:
> Also have you considered creating an extension that just adds a tag like `[HDR-Toggle] True`. Open to interpretation of course. I suppose considering you're adding c++ dependencies that it may not work, but an extension would probably be a better bet than a merge into the main repo

If there ends up being some reason why this PR never gets merged, then I would absolutely look into writing an extension that accomplishes the same goal. Because this feature is an essential part of improving my gaming experience and I believe it would be valuable for other people too. But I'll cross that bridge if I come to it. Speaking of which, @JosefNemec is there anything else that concerns you about this PR?

source/Playnite.DesktopApp/ViewModels/GameEditViewModel.cs
```diff
@@ -174,6 +174,8 @@ public bool IsBackgroundTooLage
 
         #endregion Database fields
 
+        public bool IsHdrSupported => HdrUtilities.IsHdrSupported();
```
JosefNemec, line 177: I don't think this is needed. Especially since people sometimes transfer libraries between PCs and they wouldn't be able to toggle this option on PCs that don't support and then use it on others that do. The checkbox should be always enabled.

source/PlayniteSDK/Models/Game.cs
```diff
@@ -260,6 +262,24 @@ public List<Guid> GenreIds
             }
         }
 
+        private bool hdr;
+        /// <summary>
+        /// Gets or sets value indicating if the game is should enable HDR on the primary monitor while playing.
+        /// </summary>
+        public bool Hdr
```
JosefNemec, line 269: Should be called something like `EnableSystemHDR` to make it clear what it actually does at first glance.

source/Playnite/Localization/LocSource.xaml
```diff
@@ -542,6 +542,9 @@ If disabled, games that use any item in any filter will be included in the view.
     <sys:String x:Key="LOCGameIsUnInstalledTitle">Not installed</sys:String>
     <sys:String x:Key="LOCGameHiddenTitle">Hidden</sys:String>
     <sys:String x:Key="LOCGameFavoriteTitle">Favorite</sys:String>
+    <sys:String x:Key="LOCGameHdrTitle">HDR</sys:String>
+    <sys:String x:Key="LOCGameHdrTooltip">If enabled, HDR will be enabled on the primary display before starting the game</sys:String>
+    <sys:String x:Key="LOCGameHdrNotSupportedTooltip">HDR is not supported on the primary display</sys:String>
```
JosefNemec, line 547: As my previous comment, I don't think this is needed.

source/Playnite/Localization/LocSource.xaml
```diff
@@ -542,6 +542,9 @@ If disabled, games that use any item in any filter will be included in the view.
     <sys:String x:Key="LOCGameIsUnInstalledTitle">Not installed</sys:String>
     <sys:String x:Key="LOCGameHiddenTitle">Hidden</sys:String>
     <sys:String x:Key="LOCGameFavoriteTitle">Favorite</sys:String>
+    <sys:String x:Key="LOCGameHdrTitle">HDR</sys:String>
```
JosefNemec, line 545: Should be described as "Enable HDR support" or something similar.

source/Playnite.DesktopApp/Windows/GameEditWindow.xaml
```diff
@@ -546,29 +546,51 @@
                     <CheckBox Grid.Column="2" Grid.Row="7" Margin="3"
                               IsChecked="{Binding EditingGame.Favorite}" />
 
-                    <Canvas Height="20" Grid.Column="0" Grid.Row="8" />
+                    <!-- HDR -->
+                    <TextBlock Grid.Column="1" Grid.Row="8" Text="{DynamicResource LOCGameHdrTitle}" />
```
JosefNemec, line 550: You need to add new RowDefinition for this new option.

source/Playnite/GamesEditor.cs
```diff
@@ -746,6 +753,23 @@ public void SetHideGames(List<Game> games, bool state)
             }
         }
 
+        public void SetHdr(Game game, bool state)
```
JosefNemec, line 756: Should be called `SetHdrSupport` or something similar.

source/Playnite/GamesEditor.cs
```diff
@@ -1386,8 +1416,14 @@ private void Controllers_Stopped(object sender, GameStoppedEventArgs args)
                 Application.Discord?.ClearPresence();
             }
 
+            if (wasHdrEnabled.HasValue)
```
JosefNemec, line 1419: 1) This will always have value after first game is started, so I don't think it makes sense to have it nullable.
2) What will happen if multiple games are started? Each with different HDR options.

source/Playnite/HdrUtilities.cs
```diff
@@ -0,0 +1,135 @@
...
+
+            (_, bool? isHdrEnabled) = GetHdrStatus(targetInfo.Value);
+            return isHdrEnabled;
+        }
+
+        /// <summary>
+        /// Enables/disables HDR on the primary display
+        /// </summary>
+        /// <param name="enable">True if enabling HDR, false if disabling HDR</param>
+        public static void EnableHdr(bool enable)
```
JosefNemec, line 49: Should be called `SetHdr`. Calling enable something with false option looks weird.

source/Playnite/GamesEditor.cs
```diff
@@ -325,6 +326,12 @@ void cancelStartup(string message)
                     {  "SelectedRomFile", startingArgs.SelectedRomFile }
                 };
 
+                wasHdrEnabled = HdrUtilities.IsHdrEnabled();
```
JosefNemec, line 329: All calls to `HdrUtilities` should be in try/catch (also logging potential failure). I understand that this should generally fail in crashy way, but I've seen people "customizing" their Windows installs to the point that more trivial things would fail in nonsensical way.

Might be worth handling it directly in HdrUtilities, not sure.

source/Playnite/HdrUtilities.cs
```diff
@@ -0,0 +1,135 @@
...
+            var paths = new DISPLAYCONFIG_PATH_INFO[pathCount];
+            var modes = new DISPLAYCONFIG_MODE_INFO[modeCount];
+
+            result = User32.QueryDisplayConfig(QUERY_DEVICE_CONFIG_FLAGS.QDC_ONLY_ACTIVE_PATHS, ref pathCount, paths, ref modeCount, modes, IntPtr.Zero);
+            if (result != WinError.ERROR_SUCCESS)
+            {
+                return null;
+            }
+
+            DISPLAYCONFIG_MODE_INFO primaryDisplayMode = modes
```
JosefNemec, line 102: Please use `var` instead where possible.

source/Playnite/MenuItems.cs
```diff
@@ -236,6 +238,16 @@ public static List<SearchItem> GetSearchGameMenuItems(Game game, MainViewModelBa
                 () => model.App.GamesEditor.ToggleHideGame(game))
                 { Icon = game.Hidden ? unHideIcon : hideIcon });
 
+            // Toggle HDR
+            if (HdrUtilities.IsHdrSupported())
```
JosefNemec, line 242: I don't think this is needed. As I mentioned in my previous comment, there is a case where a user might want to change this on a PC that doesn't support HDR.

source/Playnite.DesktopApp/ViewModels/GameEditViewModel.cs
lscholte: Hmm. This is an interesting point and I see where you are coming from. The counterpoint to this (and why I implemented it this way) is that it may make users believe that they have access to features that their monitors don't actually support. I want to think about this a bit more.

source/Playnite/HdrUtilities.cs
lscholte: Will do. Although this might be something you wish to clarify in coding conventions for this repo. There are counterpoints to this, specifically that explicit usage of the type may be better if the right-hand side of the `=` doesn't provide a clear indication on the type.

source/Playnite.DesktopApp/ViewModels/GameEditViewModel.cs
JosefNemec: I'm pretty a user would know if they have HDR support available on their system prior to actually using Playnite. It would be very weird use case if Playnite would actually introduce them to HDR being available to them.

JosefNemec: I think usable solution would be adding some indicator next to this option saying that HDR support was detected or not detected on this system.

source/Playnite/GamesEditor.cs
lscholte: Regarding 2, my feature basically works in the following ways:

Case 1: Game's HDR toggle is on and system HDR is currently off. Then upon launching the game, system HDR is turned on. And upon exiting the game system, HDR is turned off

Case 2: Game's HDR toggle is on and system HDR is currently on. Then upon launching the game, nothing new happens. And upon exiting the game, nothing new happens (system HDR continues to be on)

Case 3: Game's HDR toggle is off and system HDR is either on or off (doesnt matter). Upon entering the game, nothing happens, and upon exiting the game nothing happens. Basically, if the game has HDR set to disabled, then the game functions the same as if I had never written this feature.



Taking the above into account, if enable HDR for game A and have it disabled for game B and then launch both of them, then HDR will only be disabled once game A exits.

There is however a corner case that you've made me think of where if HDR is set on for both game A and B. If you launch A followed by B and then kill A, it will most likely disable system HDR even though game B still wants it. But I don't think this would happen if you launched A then B then killed B (see case 2 above)

source/PlayniteSDK/Models/Game.cs
lscholte: Done

source/Playnite.DesktopApp/ViewModels/GameEditViewModel.cs
lscholte: Ok so I have tweaked this so that the `IsHdrSupported` check still exists, but its only purpose here is to enable me to show additional info in the tooltip:
image

Likewise, this is what it would look like if my primary monitor does support HDR:
image

source/Playnite/Localization/LocSource.xaml
lscholte: I've kept this. See https://github.com/JosefNemec/Playnite/pull/3290#discussion_r1105450181 for why

source/Playnite/Localization/LocSource.xaml
lscholte: Done

source/Playnite.DesktopApp/Windows/GameEditWindow.xaml
lscholte: Done (and surprised I missed this the first time)

source/Playnite/GamesEditor.cs
lscholte: I called it `SetHdrEnabled` to match the `IsHdrEnabled`

source/Playnite/GamesEditor.cs
lscholte: Point number 1 has been addressed. I need to think more about point number 2 still

source/Playnite/GamesEditor.cs
lscholte: Added error logging into `HdrUtilities.cs` and I also added try/catches around the public methods in `HdrUtilities` so that anywhere that calls it doesn't need to do any exception handling. If something weird happens, it would just be treated as if HDR is not supported/enabled

source/Playnite/HdrUtilities.cs
lscholte: done

source/Playnite/MenuItems.cs
lscholte: Took this a step further and outright removed this menu item since it was a mistake that I still had it. Therefore, desktop mode can only toggle the HDR feature through the advanced game options dialog.

Just to make sure it is clear now, the only menu item that remains is the menu item for a game in fullscreen mode. And for that item, it only appears if the primary display supports HDR. Do you want me to make it so that the item always appears?

source/Playnite/GamesEditor.cs
lscholte: done

source/Playnite/GamesEditor.cs
lscholte: Point number 2 is now addressed as well. Essentially, I will only get the system's current HDR state if there are no games requiring HDR that are currently playing. And I will only reset the system's HDR state back to its original state once there are no more games being played that require HDR.

source/PlayniteSDK/Models/Game.cs
```diff
@@ -177,7 +177,9 @@ public enum GameField : int
         ///
         LastSizeScanDate = 90,
         ///
-        RecentActivity = 91
+        RecentActivity = 91,
+        ///
+        Hdr = 92
```
JosefNemec, line 182: Should be EnableSystemHdr since that's how the original property is named.

lscholte: Done

source/Playnite/HdrUtilities.cs
```diff
@@ -0,0 +1,166 @@
...
+                    logger.Error("Failed to retrieve primary display target info");
+                    return false;
+                }
+
+                (bool isHdrSupported, _) = GetHdrStatus(targetInfo.Value);
+                return isHdrSupported;
+            }
+            catch (Exception e) when (!PlayniteEnvironment.ThrowAllErrors)
+            {
+                logger.Error(string.Format("Failed to check if HDR is supported: {0}", e.Message));
```
JosefNemec, line 36: I'm really sorry that I missed this, it will be the last thing to change, I swear.

To log an exception, do it this way instead:
```
logger.Error(e, "Failed to check if HDR is supported.");
```

It will automatically include formatted exception info together with the message.

lscholte:
> Sorry to throw this in, but making in all depending on the primary display is not good in my opinion. Let me explain, it is a general thing:
> 
> Windows has the bad behaviour to launch everything on the primary display if opened for the first time after boot and not where this was opened from, which is quite annoying.
> 
> Also, when i'm using a EDID Emulator to play my games on the TV i'm forced to make my TV the primary display first for everything after, because i'm not capable of telling a game beforehand to launch on a specific display to my knowledge. There are some games that support switching the target display in their display options, but in every instance you can't change this setting with something like a start parameter. I mean it should be possible on DX12 games, because it made "exclusive fullscreen" obsolete.
> 
> ... Because if you for example use playnite.fullscreen, which i can tell to launch on a secondary, i am not able to also tell the game to launch on that same monitor. Being able to choose a display would also eliminate the problem of making it primary first, not being primary would make this option fail if the other screens are non-HDR displays as far as i understand. It would also be good, if there was an option to enable, to switch the primary display depending on the target display in playnite.fullscreen like steam can do.
> 
> My use case is. When i want to use the computer for desktop things during my tv gaming sessions, like google something, everything i open launches on the TV. If somehow windows would have a better behaviour in this regard a situation where i could tell my games to launch on the same screen as the playnite fullscreen app..., that would be awesome.
> 
> Just being able to use my smaller screens for desktop things while having playnite on my TV. It's like having the PC running while playing a game on the console.
> 
> If i overlooked something that would solve this my problem, feel free to correct me.

So I do agree with you that tying the HDR feature to the primary display is a limitation. If you reference https://github.com/JosefNemec/Playnite/pull/3290#issuecomment-1425127828 I have proposed some alternatives to allow for the flexibility that you want.

That being said, I am strongly in favour of moving forward with my current approach because for the people in the similar situation as me, they can start using this feature as I have implemented it. For the people in a situation like you, they can choose to not use the feature or use it in the situations where they do want HDR on the primary display. But the point is that nobody is forced to use this feature if it exists. In a future enhancement, we could explore providing more flexibility, but a preliminary investigation I did on this indicated there might be some technical challenges I would need to overcome to implement it the way I described in that comment I linked.

Gammler:
> > Sorry to throw this in, but making in all depending on the primary display is not good in my opinion. Let me explain, it is a general thing:
> > Windows has the bad behaviour to launch everything on the primary display if opened for the first time after boot and not where this was opened from, which is quite annoying.
> > Also, when i'm using a EDID Emulator to play my games on the TV i'm forced to make my TV the primary display first for everything after, because i'm not capable of telling a game beforehand to launch on a specific display to my knowledge. There are some games that support switching the target display in their display options, but in every instance you can't change this setting with something like a start parameter. I mean it should be possible on DX12 games, because it made "exclusive fullscreen" obsolete.
> > ... Because if you for example use playnite.fullscreen, which i can tell to launch on a secondary, i am not able to also tell the game to launch on that same monitor. Being able to choose a display would also eliminate the problem of making it primary first, not being primary would make this option fail if the other screens are non-HDR displays as far as i understand. It would also be good, if there was an option to enable, to switch the primary display depending on the target display in playnite.fullscreen like steam can do.
> > My use case is. When i want to use the computer for desktop things during my tv gaming sessions, like google something, everything i open launches on the TV. If somehow windows would have a better behaviour in this regard a situation where i could tell my games to launch on the same screen as the playnite fullscreen app..., that would be awesome.
> > Just being able to use my smaller screens for desktop things while having playnite on my TV. It's like having the PC running while playing a game on the console.
> > If i overlooked something that would solve this my problem, feel free to correct me.
> 
> So I do agree with you that tying the HDR feature to the primary display is a limitation. If you reference [#3290 (comment)](https://github.com/JosefNemec/Playnite/pull/3290#issuecomment-1425127828) I have proposed some alternatives to allow for the flexibility that you want.
> 
> That being said, I am strongly in favour of moving forward with my current approach because for the people in the similar situation as me, they can start using this feature as I have implemented it. For the people in a situation like you, they can choose to not use the feature or use it in the situations where they do want HDR on the primary display. But the point is that nobody is forced to use this feature if it exists. In a future enhancement, we could explore providing more flexibility, but a preliminary investigation I did on this indicated there might be some technical challenges I would need to overcome to implement it the way I described in that comment I linked.

Sorry i missed this, already a long thread. Then just see it as arguments for the follow up PR.

source/Playnite/HdrUtilities.cs
lscholte: Done

JosefNemec (approved):

PR #3315, opened 2023-02-22
Allow users to upload a custom launch video for fullscreen mode

lscholte:
I have verified that:

**Description**
Note: I am not aware of a feature request on Github for this, but I can create one if needed.

This allows users to set a custom launch video when launching fullscreen Playnite. This will allow users to create a more console-like experience if they desire.

See https://drive.google.com/file/d/1Tz0OFakTwRsSJOGSaef78xnfjFu03JWx/view?usp=sharing for a demo.  The video files I tested with were pulled from https://playnite.link/forum/thread-686.html, so thank you to whoever created those. Although they aren't part of my PR. I was just testing with them.

**Implementation Notes**
Upon saving out of the settings page, the uploaded video will be saved to a specific location in the playnite's config location. When launching fullscreen mode, the launch video is retrieved from that location.

There are some modifications to the Playnite initialization code. In particular the addition of some `await` calls. In doing so, I moved a couple of warning dialogs (specifically the administrator permission warning and the Nahilic services warning) because with the addition of awaits, those warnings would have appeared after the library updating is completed which is a non-deterministic time for it to appear and may randomly interrupt a user. Moving it ensures that it shows up immediately once the main Playnite window opens.

Video is being played by a MediaElement in WPF. I am not actually sure what formats this supports, but I know it has been working with MP4 for me.

**Outstanding Issues**
I have realized that Windows Media Pack and/or Windows Media Player (which are optional features installed with Windows) must be installed for this to work correctly. I'm going to explore this a bit more and ensure graceful handling for when this feature is not present. The current behaviour in this situation is as if you uploaded an invalid video, which is effectively the same as clearing the video (shown in the demo). There are certain installations of Windows (namely "N" versions of Windows) that don't have this installed by default, so I think it might be worth adding a note on the settings page about this dependency.

Would probably be good for this to also be included in the backup/restore feature, but currently it is not.

**Testing performed**
* Tested with different intro videos -- some with sound, some without sound
* Tested with launch video enabled but no video is uploaded
* Tested with launch video uploaded but launch video is disabled
* Tested with and without the splash screen enabled (I intend to use without the splash screen)
* Tested that Playnite's fullscreen audio is not playing alongside the launch video
*  * This is a big reason why I even went out of my way to do this feature. Because otherwise with some clever scripting, you could accomplish this. Or could probably do the same with an extension to get a similar UI to what I built, but I'd expect the same sound problem.

### DIFF
--- a/source/Playnite.DesktopApp/ViewModels/SettingsViewModel.cs
+++ b/source/Playnite.DesktopApp/ViewModels/SettingsViewModel.cs
@@ -349,6 +349,23 @@ namespace Playnite.DesktopApp.ViewModels
             });
         }
 
+        public RelayCommand UploadFullscreenIntroVideoCommand
+        {
+            get => new RelayCommand(() =>
+            {
+                string filePath = Dialogs.SelectFile("MP4 files (*.mp4)|*.mp4|All files (*.*)|*.*");
+                if (!filePath.IsNullOrEmpty())
+                {
+                    Settings.TemporaryFullscreenIntroUri = new Uri(filePath);
+                }
+            });
+        }
+
+        public RelayCommand ClearIntroVideoCommand
+        {
+            get => new RelayCommand(() => Settings.TemporaryFullscreenIntroUri = null);
+        }
+
         #endregion Commands
 
         public object DateTimeFormatAddedExample =>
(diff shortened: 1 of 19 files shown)


source/Playnite.FullscreenApp/FullscreenApplication.cs
```diff
@@ -94,7 +98,6 @@ public override bool Startup()
             SendUsageDataAsync();
 #pragma warning restore CS4014
             ProcessArguments();
-            InitializeAudio();
```
lscholte, line 97: Just thought I'd point out that this is specifically being moved because `OpenMainViewAsync` is an async method but it's not being awaited. So the problem I'd end up having is that the intro video would start playing which is awaited inside `OpenMainViewAsync`, but since `OpenMainViewAsync` itself is not awaited, the rest of the app initialization including `InitializeAudio` would occur. This end result of this is that I'd hear audio from the intro video plus the general playnite audio simultaneously, which was not desired.

source/Playnite.DesktopApp/ViewModels/SettingsViewModel.cs
```diff
@@ -349,6 +350,23 @@ public RelayCommand ResetDateTimeFormatLastPlayedCommand
             });
         }
 
+        public RelayCommand UploadFullscreenIntroVideoCommand
+        {
+            get => new RelayCommand(() =>
+            {
+                string filePath = Dialogs.SelectFile("MP4 files (*.mp4)|*.mp4|All files (*.*)|*.*");
```
lscholte, line 357: "MP4 files (*.mp4)" and "All files (*.*)" are technically user-facing strings and arguably could be translatable. Let me know if that is something worth separating out. I've done it in the past, so it's no big deal.

JosefNemec:

PR #3316, opened 2023-02-23
Add support to display game Added Date in game details

darklinkpower:
I have verified that:

image

image

image

### DIFF
--- a/source/Playnite/Settings/DetailsVisibilitySettings.cs
+++ b/source/Playnite/Settings/DetailsVisibilitySettings.cs
@@ -83,7 +83,7 @@ namespace Playnite
             }
         }
 
-        private bool added = true;
+        private bool added = false;
         public bool Added
         {
             get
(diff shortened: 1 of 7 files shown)


source/Playnite/Settings/DetailsVisibilitySettings.cs
```diff
@@ -83,6 +83,21 @@ public bool LastPlayed
             }
         }
 
+        private bool added = true;
```
JosefNemec, line 86: I don't think this should be enabled by default. It's obviously not that important since nobody noticed until now that we don't even have an option for this :)

darklinkpower: Personally I would have preferred to make it enabled so people know of its existence but changed in https://github.com/JosefNemec/Playnite/pull/3316/commits/5ba93435980107d5ddeefb8b4db307c415bc1a62

JosefNemec (approved):

Issue #3312, opened 2023-02-22
Converting playtime from hh:mm to dd:hh:mm


karlroush:
### Feature description

Currently the playtime displayed in the Library is formatted as hours:minutes (e.g. 25h 10min).
I am proposing it be changed to days:hours:minutes (e.g. 1d 1h 10min).

This can be accomplished by changing "PlayTimeToStringConverter.cs". I'd love to add my code to contribute, but I don't know how to make a pull request for the modifications to this file.
https://github.com/JosefNemec/Playnite/blob/3dcfe90f7a2b307769496c3a94622ed0e9ba51ef/source/Playnite/Converters/PlayTimeToStringConverter.cs

add on line 22 
`private static string LOCPlayedDays;`

Then replace lines 57-64 
````
var hours = minutes / 60;
if (hours <24)
{
	return string.Format(LOCPlayedHours, hours, minutes - (hours * 60));
}

var days = (hours-hours%24)/24; //eg (49-49%24)/24 -> (49-1)/24 -> 2 days
return string.Format(LOCPlayedDays, hours%24, minutes - ((hours%24) * 60))
````

I'm not sure how this impacts string.Format though since I'm doing this just off the top of my head.

darklinkpower:
If implemented, the modification should be configurable in settings instead of changing and making it the default behavior so it would require more changes. I think it should go below the settings here in my opinion:

![image](https://user-images.githubusercontent.com/1389286/220512778-432ed94c-0a18-4fc2-97dc-f7c075dc05e5.png)

JosefNemec:
Done in 10.10

Issue #1729, opened 2020-04-09
Localization - plural forms


LilianaMoore:
Please add _'one, few, many, other'_ plural forms for variables used in strings such as 
> Are you sure you want to remove `{0}` games?

JosefNemec:
I'll be honest, I don't understand what would this be used for. We already have a different variation for a single game.

LilianaMoore:
Many languages have two grammatical forms for singular and plural phrases e.g. _"1 game"_ vs _"2 games"_. However, some languages have several forms (e.g. Slavic family; Polish, Russian, Czech, etc) and the handling of plural forms differ widely between the language families. For example in Polish, we use _"game"_ (gra) this way:
```
1 gra
2-4,22-24,32-34... gry
0, 5~19, 100, 1000... gier
0.0-1.5, 10.0, 100.0... gry
```
and so on. 
(It is worth noting that using strings like _"game(s)_" look unnatural and should be avoided.)

![plurals](https://user-images.githubusercontent.com/63408277/78930685-ce255600-7aa4-11ea-9415-4306112ff60e.png)

https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html
http://cldr.unicode.org/index/cldr-spec/plural-rules
https://www.soluling.com/Help/NET/Index.htm#Patterns

PR #3317, opened 2023-02-23
Support to indicate number of days played in game time played format

darklinkpower:
I have verified that:

Closes https://github.com/JosefNemec/Playnite/issues/3312

Setting is disabled by default

### Screenshots

image

image

image

image

image

image

### DIFF
--- a/source/Playnite/Converters/PlayTimeToStringConverter.cs
+++ b/source/Playnite/Converters/PlayTimeToStringConverter.cs
@@ -19,6 +19,7 @@ namespace Playnite.Converters
         private static string LOCPlayedSeconds;
         private static string LOCPlayedMinutes;
         private static string LOCPlayedHours;
+        private static string LOCPlayedDays;
 
         private static void CacheStrings()
         {
@@ -32,6 +33,7 @@ namespace Playnite.Converters
             LOCPlayedSeconds = ResourceProvider.GetString("LOCPlayedSeconds");
             LOCPlayedMinutes = ResourceProvider.GetString("LOCPlayedMinutes");
             LOCPlayedHours = ResourceProvider.GetString("LOCPlayedHours");
+            LOCPlayedDays = ResourceProvider.GetString("LOCPlayedDays");
         }
 
         public object Convert(object value, Type targetType, object parameter, System.Globalization.CultureInfo culture)
@@ -61,6 +63,15 @@ namespace Playnite.Converters
             }
 
             var hours = minutes / 60;
+            if (parameter is bool formatToDays && formatToDays && hours >= 24)
+            {
+                var days = hours / 24;
+                var remainingHours = hours % 24;
+                var remainingMinutes = minutes % 60;
+
+                return string.Format(LOCPlayedDays, days, remainingHours, remainingMinutes);
+            }
+
             return string.Format(LOCPlayedHours, hours, minutes - (hours * 60));
         }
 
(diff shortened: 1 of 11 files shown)


Pixxel123:
It may be nice to have the days indicator account for singular and plural by showing "days" instead of just day if the number is equal to 1. I would say this goes for 0 as well, so 0 days, 2 days, 3 days, etc.

lscholte:
The inconsistency of using "days" as a full word vs "h" and "m" for hours and minutes bothers me slightly. Given the lack of real estate where this is going to appear, I don't think using "hours" and "minutes" (or "mins") would make sense. So the alternative could be just using "d" instead of days. As a side benefit, it would address https://github.com/JosefNemec/Playnite/pull/3317#issuecomment-1443414832 too.

It's food for thought at least.

darklinkpower:
I've changed it to use `d` only instead of `days`. In my opinion better plural support should wait until https://github.com/JosefNemec/Playnite/issues/1729 is done

source/Playnite/Converters/PlayTimeToStringConverter.cs
```diff
@@ -61,6 +63,15 @@ public object Convert(object value, Type targetType, object parameter, System.Gl
             }
 
             var hours = minutes / 60;
+            if (parameter is bool formatToDays && formatToDays && hours >= 24)
```
JosefNemec, line 66: Please add unit test for this.

JosefNemec:
As far as formatting depending on number of days and various plural forms, that will definitely have to wait for #1729 which will add support for Project Fluent localization strings.

source/Playnite/Converters/PlayTimeToStringConverter.cs
darklinkpower: Done in https://github.com/JosefNemec/Playnite/pull/3317/commits/9a9c7a30cf2694a9471f371e26b456f70502d085

source/Tests/Playnite.Tests/Converters/PlayTimeToStringConverterTests.cs
```diff
@@ -0,0 +1,50 @@
...
+    [TestFixture]
+    public class PlayTimeToStringConverterTests
+    {
+        private static string Convert(object value, object formatToDaysParameter = null)
+        {
+            return PlayTimeToStringConverter.Instance.Convert(value, typeof(string), formatToDaysParameter, CultureInfo.CurrentCulture) as string;
+        }
+
+        [Test]
+        public void ConvertLocalizationStringsTest()
```
JosefNemec, line 23: This tests doesn't actually test anything because localization resources are not working in test projects. You are always comparing placeholder string with placeholder string. I have fixed that in devel so localization resource will work now.

Even if those string have worked however, you would be comparing the original string mask so the test would always fail.

The test should test against final strings, like this:
```
Assert.AreEqual("25h 20m", Convert(time, false))
```

darklinkpower: The test actually succeeded. I found after initially creating the test that localization resources didn't work, it just returned the default `<!LOCStringName>` string that gets obtained when the resource is not available. For this reason, both the Localization string in the test and what was obtained in the converter was that same not-found string that made them match.

Test reimplemented in https://github.com/JosefNemec/Playnite/pull/3317/commits/77e2a186d72af1c6c22bdfceb56337bb215e23bf

JosefNemec (approved):

Issue #2983, opened 2022-07-10
"Are you sure you want to remove this game?" Should display the name of the game.


falcorr:
### Feature description

For such an important action, the name of the game(s) being removed should be displayed at the same time. This has been especially concerning given #2982.

Would also be good to keep in mind for any other significant actions.

lscholte:
@JosefNemec You should be able to close this issue now.

PR #3318, opened 2023-02-24
Show game name in the confirmation dialog before removing game

lscholte:
I have verified that:

**Description**
Shows the game name in the confirmation dialog before it is removed. This addresses https://github.com/JosefNemec/Playnite/issues/2983
image

**Testing Performed**
Tested with a steam game and an emulator game (they use different strings)

### DIFF
--- a/source/Playnite/GamesEditor.cs
+++ b/source/Playnite/GamesEditor.cs
@@ -859,7 +859,7 @@ namespace Playnite
                 {
                     addToExclusionList = true;
                 }
-                else if (result == options[2])
+                else if (result == null || result == options[2])
                 {
                     return;
                 }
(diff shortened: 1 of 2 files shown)


source/Playnite/GamesEditor.cs
```diff
@@ -850,15 +851,15 @@ public void RemoveGame(Game game)
...
+                    string.Format(resources.GetString("LOCGameRemoveAskMessageIgnoreOption"), game.Name),
                     "LOCGameRemoveAskTitle",
                     MessageBoxImage.Question,
                     options);
                 if (result == options[0])
                 {
                     addToExclusionList = true;
                 }
-                else if (result == options[2])
+                else if (result == null || result == options[2])
```
lscholte, line 862: FYI this is something I noticed where if you were to close the game via something other than the action buttons on the dialog (e.g alt-tab and close the dialog), then the logic would proceed to remove the game even though this should more naturally cancel the operation (i.e not remove the game).

JosefNemec (approved):

PR #3325, opened 2023-02-28
Updated .NET Framework download link

Jeshibu:
Existing one's broken

I have verified that:


JosefNemec:
The link should just point to the latest .NET Framework version since most people will have newer version than 4.6.2 installed already and they are all backwards compatible.

JosefNemec:
Now that I think about it, it would be best to just remove any mention of .NET Framework download. I think the only case where you wouldn't have compatible version installed is fresh Windows 7 install. And in that case Playnite will prompt you to install .NET Framework anyways.

PR #3331, opened 2023-03-05
Fixes a bug where fullscreen audio does not play upon resuming from sleep/hibernate

lscholte:
I have verified that:

**Description**
I have been noticing a bug in fullscreen mode where the audio stops playing when you resume a session after putting the computer to sleep or hibernate. Another user has also described this issue in https://www.reddit.com/r/playnite/comments/103k592/audio_issue/

This fix detects the resumption of a user's session and simply re-initializes the audio when that happens.

**Outstanding Issues**
I'm not aware of anything.

**Testing Performed**
I have so far been testing on a Surface Laptop, but I have 2 other computers that I plan to test on later. Testing has included
* Locking and unlocking the computer -- this restarts the audio
* Sleeping the computer and then resuming -- this restarts the audio
* Hibernating the computer and then resuming -- this restarts the audio
* Switching to a different user and then switching back -- this restarts the audio
* Minimizing playnite and then re-maximizing it -- this does not restart the audio, but this also never had a problem to begin with since the audio would just naturally resume (if background audio was disabled when not focused) or continue playing always (if background audio was enabled even when not focused)

### DIFF
--- a/source/Playnite.FullscreenApp/FullscreenApplication.cs
+++ b/source/Playnite.FullscreenApp/FullscreenApplication.cs
@@ -1,4 +1,5 @@
-﻿using Playnite.API;
+﻿using Microsoft.Win32;
+using Playnite.API;
 using Playnite.Audio;
 using Playnite.Common;
 using Playnite.Controllers;
@@ -96,6 +97,7 @@ namespace Playnite.FullscreenApp
             ProcessArguments();
             InitializeAudio();
             PropertyChanged += FullscreenApplication_PropertyChanged;
+            SystemEvents.SessionSwitch += HandleAudioAfterSessionSwitch;
             return true;
         }
 
@@ -210,6 +212,7 @@ namespace Playnite.FullscreenApp
         {
             StopBackgroundSound();
             Audio?.Dispose();
+            SystemEvents.SessionSwitch -= HandleAudioAfterSessionSwitch;
             base.ReleaseResources(releaseCefSharp);
         }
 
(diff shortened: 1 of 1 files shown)


JosefNemec:
Why did you close this?

lscholte:
> Why did you close this?

Sorry I probably should have left a comment. I need to do more work on this. It seems to work for one of my computers but not one of my others (although it wasn't any worse either). I want to look at it a bit further and then re-open if/when I get more confident that it is working correctly

JosefNemec:
Btw we are switching to a completely different audio library in P11 so it might not be an issue then.

lscholte:
I will probably re-open the PR again in the next day or two. I have an alternative way of checking that the audio stopped rather than using anything to do with the state of Windows. Instead, we can listen for a `PlaybackStopped` event that gets raised and it specifically sets an exception property in the event arguments if it was an unexpected stoppage, and this does seem to be triggered when I resume from sleep/hibernate. The problem is that the audio reinitialization that I was (and still am doing) doesn't always work and I've got no idea why. My theory at the moment could be that it's a timing issue with Windows still being resumed from sleep state. I guess to prove that I could play around with some arbitrary sleeps, but I've not had the opportunity to really dive into what is actually happening.

When I reopen this PR, I will make the note that I don't think the changes should cause regressions, but I'm also not convinced it will solve the issue 100% of the time either. You may then do whatever you wish with the PR. If you are switching audio libraries in the future, it is quite possible that could resolve the issue more fully.

Issue #3300, opened 2023-02-12
Switching audio devices while Playnite Fullscreen is focused does not work


Super1337Boy:
### Bug Description

I have button combinations set up to switch between audio devices on the fly from my controller (using Controller Companion and NirCmd), but switching devices this way doesn't work in Playnite Fullscreen.

The behavior differs depending in which audio API is used:

WASAPI: Switching devices doesn't work at all. The sound stays with the original device when switching.
DirectSound: Sound disappears when switching, and restarting Playnite with the right device active is necessary to brings the sound back.

It's worth noting that the behavior is different if I take focus away from Playnite and switch between devices from the taskbar. When doing this, switching devices work fine while using DirectSound, but still not with WASAPI.

Is there anything that can be done to make switching devices work on the fly?

### Diagnostics ID

1e325b2d-43e5-4e73-8f1c-70173decd988

Super1337Boy:
Could this issue by any chance potentially have a similar solution to the fixes in this PR: #3331 

I'm sorry if I'm way off here (I'm not a programmer), but maybe the issue I'm having could be fixed by re-initializing audio when the audio device is changed?

JosefNemec:
No idea. I'm not looking into these audio issue right now, because I'm switching audio library completely in Playnite 11.

lscholte:
I re-opened the PR under https://github.com/JosefNemec/Playnite/pull/3335. That said, I'm not that confident it will fix your issue as I have tried plugging/unplugging my headphones and the audio is just broken. I guess we'll see how audio ends up behaving with Playnite 11.

PR #3335, opened 2023-03-11
Potentially fixes a bug where fullscreen audio does not play upon res…

lscholte:
I have verified that:

I originally had this open under https://github.com/JosefNemec/Playnite/pull/3331

Description
I have been noticing a bug in fullscreen mode where the audio stops playing when you resume a session after putting the computer to sleep or hibernate. Another user has also described this issue in https://www.reddit.com/r/playnite/comments/103k592/audio_issue/

This fix detects the that audio playback has stopped unexpectedly and simply re-initializes the audio when that happens. 

Outstanding Issues
I believe this is PR is not necessarily guaranteed to fix the problem. It works most of the time for me, particularly when resuming from a hibernation. But resuming from sleep tends to be hit or miss. That said, this logic should only ever kick in if the audio wasn't working anyway, so I figure the re-initialization is worth a shot. I am aware that you intend to change audio libraries in Playnite 11, so it's up to you what you want to do with this PR.

Testing Performed
I have so far been testing on a Surface Laptop, but I have 2 other computers that I plan to test on later. Testing has included

Locking and unlocking the computer -- audio simply resumes playing normally
Sleeping the computer and then resuming -- audio crashes and then restarts (usually but doesn't always restart successfully)
Hibernating the computer and then resuming -- audio crashes and then restarts
Switching to a different user and then switching back -- this restarts the audio
Minimizing playnite and then re-maximizing it -- audio simply resumes playing normally
Plugging/unplugging headphones -- audio stops playing and does not restart. This indicates that this probably will not resolve https://github.com/JosefNemec/Playnite/issues/3300

### DIFF
--- a/source/Playnite.FullscreenApp/FullscreenApplication.cs
+++ b/source/Playnite.FullscreenApp/FullscreenApplication.cs
@@ -1,4 +1,5 @@
-﻿using Playnite.API;
+﻿using NAudio.Wave;
+using Playnite.API;
 using Playnite.Audio;
 using Playnite.Common;
 using Playnite.Controllers;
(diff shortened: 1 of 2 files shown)


cnapolit:
Observed this issue as well: https://github.com/JosefNemec/Playnite/issues/3391
Closed the ticket given there is already a PR for this.

JosefNemec:

PR #3541, opened 2023-09-04
Adds checkbox for uncensoring images

sapioit:
While I know that new pull requests aren't being accepted, please leave this open, so I could see the changes I proposed and make another pull request after v11 if you don't solve this problem yourselves before then. 


JosefNemec:
Also the reason why PRs are not being accepted right now is because it will be basically impossible to merge them after P11 changes are public due to amount of refactoring done to the code base. So having open PR is kind of pointless until that's done.

I definitely appreciate people wanting to contribute but Playnite is simply in a state right now where I don't accept code contributions until P11 is in beta state.

sapioit:
Thank you! If this weren't already implemented, I would have wanted to still have this here, to have a rough idea of what the goal was and how I got it done, so I can do it again after the refactoring. Thank you for implementing it already! I look forward to using that option! I'm looking forward to having the v10.19 ready for download, or have the v11 beta ready for download instead, whichever comes first. 

Have a nice day!

PR #3571, opened 2023-09-30
Keyboard launcher results improvements

### LINKED ISSUE #3570: Improve keyboard launcher search results by ordering with weight ranking system
### Feature description

Currently the keyboard launcher results are ordered by name. Ideally it should use a weigth ranking system to provide the more relevant results on top. Here is an example of why this is currently not ideal, with the search term `dusk`. The list includes a game with the exact same name but it's down in the list, ideally it should be placed on top as it being in reality the best match.

As ideas, the Levenshtein distance or JaroWinkler similarity algorithms could be used for this. Here are implementations of them: https://github.com/darklinkpower/PlayniteExtensionsCollection/blob/5c7dceb7dba2c123376f829a33b1f839b0897897/source/Common/PluginsCommon/Extensions/StringExtensions.cs#L396-L563

I don't have benchmarks at hand but they are extremely fast and worked better from all the implementations I found.

### Screenshots

6X6jOiinxd

darklinkpower:

Sorry I didn't know that PRs are closed. Opening just in case it could be considered but if there's any issue please feel free to close it :(

---

Closes #3570

Changes:
1. **Improve game results ordering:** Improves game order sorting of keyboard launcher game results to provide more relevant results on top by sorting by [Levenshtein Distance algorithm](https://en.wikipedia.org/wiki/Levenshtein_distance) to provide more relevant results on top. 
2. Subsequently sorts by name and install status (Which is useful if there are duplicates from multiple plugins)
3. **Fuzzy matching:** Implement fuzzy matching for filtering by using [Jaro Wrenklin distance algorithm](https://en.wikipedia.org/wiki/Jaro%E2%80%93Winkler_distance)

For order sorting I experimented with [Jaro Wrenklin distance algorithm](https://en.wikipedia.org/wiki/Jaro%E2%80%93Winkler_distance) with different threshold minimum weight values (Used to determine if a string should be considered a match) and [Levenshtein Distance algorithm](https://en.wikipedia.org/wiki/Levenshtein_distance), which determines how close is a given string from another and found the latter to be much better in sorting results for this use case.

For fuzzy matching, Jaro Wrenklin works much better in conjunction with all the other filtering methods already being used from my testing.

### Sorting changes examples
---
**Example 1**

Before:
FlpOX0oIIy

After:
image

**Example 2**

Before:
image

After:
image

**Example 3**

Before:
image

After:
image

### Fuzzy matching examples
---

Before:
image

After:
image

### DIFF
--- a/source/Playnite/Common/Extensions/StringExtensions.cs
+++ b/source/Playnite/Common/Extensions/StringExtensions.cs
@@ -11,9 +11,25 @@ using System.Web;
 
 namespace System
 {
+    class CaseInsensitiveCharComparer : EqualityComparer<char>
+    {
+        public override bool Equals(char x, char y)
+        {
+            return char.ToUpperInvariant(x) == char.ToUpperInvariant(y);
+        }
+
+        public override int GetHashCode(char obj)
+        {
+            return char.ToUpperInvariant(obj).GetHashCode();
+        }
+    }
+
     public static class StringExtensions
     {
         private static readonly CultureInfo enUSCultInfo = new CultureInfo("en-US", false);
+        private const double defaultWinklerWeightThreshold = 0.7; //Winkler's paper used a default value of 0.7
+        private const int winklerNumChars = 4; //Size of the prefix to be considered by the Winkler modification.
+        private static readonly EqualityComparer<char> charCaseInsensitiveComparer = new CaseInsensitiveCharComparer();
 
         public static string MD5(this string s)
         {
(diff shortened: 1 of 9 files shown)


darklinkpower:
Update

Also added fuzzy matching using [Jaro Wrenklin distance algorithm](https://en.wikipedia.org/wiki/Jaro%E2%80%93Winkler_distance)

Before:
image

After:
image

JosefNemec:
I would maybe consider this for 10.20, but there are no tests. Please add tests to check that this doesn't break things like acronym search.

darklinkpower:
Sorry but I'm not understanding and apologize if I'm wrong but shouldn't passing the existing tests mean nothing is breaking?

image

If not what tests are you thinking of?

P.S. There are some existing broken tests that I fixed in https://github.com/JosefNemec/Playnite/pull/3571/commits/77b0825a335ea8c8deb461f95d57029222e16530

JosefNemec:
I'm stupid, I forgot that we already have tests for acronym search, I'm sorry.

Though it would be good to have tests specifically to check that this new ordering works.

darklinkpower:
I've added a filtering test in https://github.com/JosefNemec/Playnite/pull/3571/commits/357cb487b6430af94b6e3a0f1fb8102ed3a4be1c, as well as tests for JeroWrenklin and Levenshtein algorithms in https://github.com/JosefNemec/Playnite/pull/3571/commits/f240b74718bd4e8067aad4e73ae8dcebd9faca79. The latter includes an ordering test that is the same as the one used during searches in `GetSearchResults()`.

Do you think that would suffice for the tests? I'm checking how to implement the test for the Search results ordering itself and it would be a mess because it requires creating a game database, a search context, a main view model (Which requires several objects in the constructors, which also require more objects in their constructors), since the main view model is an abstract class it would require adding a reference to the Desktop project, etc. It's complicated. Here's an image of the start of this rabbit hole:

image

In my opinion currently the tests in the PR should suffice and perhaps in Playnite 11 more tests could be added after the refactor, for this and more things.

JosefNemec:
You don't need to replicate the entire setup. You can just run the query that starts with that `Where(g => GameFilter(g, searchTerm, args.GameFilterSettings, true))` on arbitrary list of games.

JosefNemec:
Generally for unit tests you want to deal with the minimal unit of code needed, which in this case would be that inner linq query. The rest of that search view model is not important for this specific ordering test.

JosefNemec:
Looking into it more, you might be right that this is properly not that critical right now. Those other tests you've added will suffice for now.

I will make formal request for proper test coverage when search rewrite is ready for P11 😉

PR #3682, opened 2024-02-06
Lenovo Legion Go controller issues adressed

lburygin:
Details here: https://www.reddit.com/r/LegionGo/comments/1afc5pc/solution_for_double_inputs_in_playnite_after/


JosefNemec:
> // there is a bug in SDL library - probably some DEVICEREMOVED events don't fire or DEVICEADDED events fire twice

Do you have link to the bug? Did you try the latest version of SDL to see if it's fixed there?

Also, why is this new option needed if this PR completely reloads all devices anyways so there shouldn't be any ghosts left because we didn't get removed event from SDL.

lburygin:
> Do you have link to the bug? 
Unfortunately, I dont have the link, I just guessed that a bug is in SDL cause nothing fixes it except reloading gamecontroller subsystem. 

>Did you try the latest version of SDL to see if it's fixed there?
I tried latest version but I had problems with interop code (Playnite crashed on launch) and didn't have time to figure out yet how to properly update it, sorry. But it's reasonable to try on new version, you are right


> Also, why is this new option needed if this PR completely reloads all devices anyways so there shouldn't be any ghosts left because we didn't get removed event from SDL.
Because there is more than one problem. In addition to SDL ghosts, there (I guessing) is an issue with Lenovo Drivers - controller being registered as new device in device manager every time it connects/disconnects, so device ID is different and if I add controller to DisabledGameControllers, after reload it will be unchecked because ID is not the same. So, if one uses Handheld Companion with XInput emulation it will have double triggers even after SDL reloading. The new setting 'fixes' it for this particular case (not ideal solution, but I think there is nothing we can do at the current moment, maybe you have some better ideas)

Sorry for my English

JosefNemec:
There shouldn't be any interop issue with updated SDL, they are not making breaking signature changes in current (2.x) release as far as I know. Just make sure you are using 32bit release dlls since Playnite is 32bit process.

PR #3697, opened 2024-02-27
Properly close Playnite before system shutdown

mammo0:
> **Code contributions (pull requests) are currently not being accepted while majority of code base is being rewritten for Playnite 11.**
> **Please wait with any pull requests after P11 is at least in beta state.**

At first excuse me, but I read the above statement too late. Even if this PR will not be accepted, I would be happy to see my changes implemented in version 11 of Playnite.

**PR description:**
With this PR the fullscreen application of Playnite gets properly closed if a system shutdown or restart is requested by the user. The current implementation only runs the Windows `shutdown` command and during the shutdown the Playnite process will be closed by the system.
Now, the Playnite application is closed before the `shutdown` command is called. This ensures especially that any custom scripts that should be run on application exit will be definitely executed.

### DIFF
--- a/source/Playnite.FullscreenApp/ViewModels/MainMenuViewModel.cs
+++ b/source/Playnite.FullscreenApp/ViewModels/MainMenuViewModel.cs
@@ -51,6 +51,12 @@ namespace Playnite.FullscreenApp.ViewModels
             this.MainModel = mainModel;
         }
 
+        private void QuitApplication()
+        {
+            MainModel.CloseView();
+            MainModel.App.Quit();
+        }
+
         public bool? OpenView()
         {
             return window.CreateAndOpenDialog(this);
@@ -64,8 +70,7 @@ namespace Playnite.FullscreenApp.ViewModels
         public void Shutdown()
         {
             Close();
-            MainModel.CloseView();
-            MainModel.App.Quit();
+            QuitApplication();
         }
 
         public void SwitchToDesktopMode()
@@ -129,17 +134,26 @@ namespace Playnite.FullscreenApp.ViewModels
                 return;
             }
 
-            if (!PlayniteEnvironment.IsDebuggerAttached)
+            // First add a custom Application.Exit event handler which gets called on application exit. This handler performs the comuter shutdown.
+            PlayniteApplication.CurrentNative.Exit += delegate
             {
-                try
+                if (!PlayniteEnvironment.IsDebuggerAttached)
                 {
-                    Computer.Shutdown();
+                    logger.Info("Shutting down computer");
+
+                    try
+                    {
+                        Computer.Shutdown();
+                    }
+                    catch (Exception e)
+                    {
+                        logger.Error(e, "Failed to shutdown the computer.");
+                    }
                 }
-                catch (Exception e)
-                {
-                    Dialogs.ShowErrorMessage(e.Message, "");
-                }
-            }
+            };
+
+            // Then quit Playnite regularly.
+            QuitApplication();
         }
 
         public void HibernateSystem()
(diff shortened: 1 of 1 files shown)


JosefNemec:
Unless I'm missing something, then this is completely unnecessary. We are already properly closing the app, including calling exit scripts and executing SDK exit events, because we are hooked into WPF's `SessionEnding` event which is called when the system is shutting down or rebooting.

I tested it myself and app exit script is properly called, even when the system is shut down via Fullscreen mode's main menu (which is calling `shutdown` command).

mammo0:
Hello,

thank you for the hint with the `SessionEnding` event. I didn't notice that you are using this.
But unfortunately this doesn't seem to work in my case.

My setup:
Normally I have a widescreen monitor connected to my PC. But I also have a connection to my TV via [Sunshine](https://github.com/LizardByte/Sunshine). As you can imagine I use the fullscreen mode of Playnite while streaming to the TV. Because my TV has another resolution as my widescreen monitor I use the app "QRes" for changing the resolution on Playnite startup and for changing it back when closing Playnite.

If I just quit Playnite everything works. But if I initiate the system shutdown through Playnite, QRes is apparently not called, because the resolution does not change.

Maybe the `SessionEnding` event is fired to late?

If I use the implementation of my PR (first close Playnite and initiate the system shutdown with the `Application.Exit` event), then my custom shutdown script works and I have the correct resolution.

JosefNemec:
Make sure you are actually waiting for that app you are running to change resolution to exit. If you are just starting an exe without waiting for it to finish, it's likely that system shuts down before it actually properly starts and does its thing.

Shutdown script is definitely properly executed according to my tests on system shutdown, so it's most likely an issue with implementation of your shutdown script.

mammo0:
> so it's most likely an issue with implementation of your shutdown script.

Actually it isn't a real script. I'm calling the QRes.exe directly. I just entered the path to the exe and the required arguments into the text box in the Playnite settings.
This should normally wait for the exe to finish? I'm not using any `&` at the end of the command or `Start-Job` of Powershell.

JosefNemec:
Use `Start-Process` with `-Wait` argument.

https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/start-process?view=powershell-5.1

mammo0:
> Use `Start-Process` with `-Wait` argument.

I will try it and report back. But I won't get to it until tomorrow.

JosefNemec:
> The application was unable to start correctly (0xc0000142). Click OK to close the application

Try specifying working directory in Start-Process call. Some apps don't start properly without it.

> Wouldn't it be better if the Playnite shutdown script is executed before the actual system shutdown command is called?

We would still have to have separate code path for cases where shutdown is not invoked from Playnite. Which happens pretty often with Desktop mode where you have it just minimized in tray when a user goes to shutdown PC manually.

mammo0:
> Try specifying working directory in Start-Process call. Some apps don't start properly without it.

Unfortunately this changes nothing. If just quit Playnite, calling the exe works without problems. The error only occurs if I request a system shutdown or restart through Playnite...

> We would still have to have separate code path for cases where shutdown is not invoked from Playnite. Which happens pretty often with Desktop mode where you have it just minimized in tray when a user goes to shutdown PC manually.

I get your point. And I don't want to remove the handling of the `SessionEnding` event. Only in the cases where the system shutdown is requested through Playnite, it should first quit probably and then call the shutdown command.

JosefNemec:
Made simpler implementation in e6e8b36cb3b0ce71fb2a2d680093006d233eb9b3, will be merged in next P10 update.

mammo0:
Thank you :)

I wasn't aware of the `QuitAndStart` method, but I think this will work. Maybe I'll find some time in the next days to test your commit.

mammo0:
Can confirm, with commit https://github.com/JosefNemec/Playnite/commit/e6e8b36cb3b0ce71fb2a2d680093006d233eb9b3 the script is executed correctly at system shutdown.

Thank you for looking into this :)

Issue #3824, opened 2024-09-17
Please add suyu emulator


henuoxi:
### Feature description

Hello, first of all, I am sorry, because the next feedback is translated by a translator, there may be some errors, first of all, why I want to add suyu emulator because the yuzu emulator has been closed by Nintendo, but now there may be people still using it.

henuoxi:
So I hope to detect the Yuzu do not delete, and then the Suyu simulator is the descendant of Yuzu, so I hope that the future version of the update can add detection Suyu simulator to fix this bug, above is my feedback to you.

henuoxi:
So I hope to detect Yuzu do not delete, and then the suyu simulator is the descendant of Yuzu, so I hope that the future version of the update can add detection of Suyu simulator, above is my feedback to you

JosefNemec:
I'm not planning to add profiles for any Yuzu fork because of scuffed the entire ecosystem around Yuzu forks is. You can always make custom profile.

Issue #3723, opened 2024-04-02
Add support for the suyu emulator


cromefire:
### Feature description

Basically [Suyu](https://suyu.dev/) is resuming development on yuzu (in a more sterile environment...) and should work basically the same except for some renaming. If you try to add it as yuzu, it won't find the launch binary.

### Screenshots

<details>

![image](https://github.com/JosefNemec/Playnite/assets/26320625/36314fe9-6aa3-4471-9513-4d9f46dbbf85)
![image](https://github.com/JosefNemec/Playnite/assets/26320625/b7d2c889-d80d-4c4a-8ee7-a4861c19c029)

</details>

JosefNemec:
You can make custom profile for emulators that don't have built-in support, including Suyu. I'll probably wait a bit before adding built-in support for it given its history.

cromefire:
Understandable. But now there's at least an issue to keep track of it

DmitriyPekar:
https://github.com/sudachi-emu/sudachi

JosefNemec:
Closing. You can configure it manually: https://api.playnite.link/docs/manual/features/emulationSupport/addingNewEmulators.html

It's unlikely we are going to add built-in support due to the entire shit show around Yuzu forks.

cromefire:
Yeah that got far more convoluted than expected...

PR #3838, opened 2024-09-27
Emulators: Add support for suyu

sounddrill31:

This is just a teeny tiny emulator profile PR so ^ shouldn't matter, right?

I know the current decision is to not add Suyu based on https://github.com/JosefNemec/Playnite/issues/3824 and https://github.com/JosefNemec/Playnite/issues/3723 but it's here if you need it.


JosefNemec:
I'm going to close this based on comment in those issues you linked. There's now also this plugin that could be used: https://playnite.link/addons.html#ExtraEmulatorProfiles

Issue #3897, opened 2024-12-25
shadps4 emulator support


Gentleman2292:
### Feature description

i would like to be able to add shadps4 games quickly from the shadps4 emulator to playnite
https://github.com/shadps4-emu/shadPS4/

vaughngx4:
Is anyone working on this? Is it possible to implement this before the Playnite 11 beta is released? Wouldn't mind taking a crack at it.

JosefNemec:
You can add any emulator you want manually without explicit support from Playnite https://api.playnite.link/docs/manual/features/emulationSupport/addingNewEmulators.html#manual-emulator-configuration

vaughngx4:
@JosefNemec shadPS4 requires a scripted scan to import games automatically. Very similar to the RPCS3 script. I've managed to add support in the source code (I've only tested the import script so far) but I've had no luck adding it manually using the "manual emulator configuration" in Playnite itself.

Any chance you can direct me to a guide on how to build from source?

vaughngx4:
I've added shadPS4 emulator support on my fork of Playnite (devel-emu_shadps4 branch)). Tested and working. At the request of @JosefNemec I shall not create a pull request until Playnite 11 beta is released.

JosefNemec:
@vaughngx4 open PR anyways please. Some other things came up and another P10 update is likely now.

vaughngx4:
Just realized that I made the PR to the wrong branch. I'm on mobile right now, I will fix it when I get home.

vaughngx4:
@JosefNemec After failing at creating a proper PR twice I shall refer to you for assistance. The ReadMe says to PR to "devel" but my modifications are made on a fork from "master". Should I moved my changes to the devel branch before PR or PR directly to master? Apologies for the mess I made... Closed both PRs myself to avoid confusion.

JosefNemec:
Just make your changes against devel then. I updated devel to be more up to date.

vaughngx4:
This issue can be closed. shadPS4 support patch has been merged.

PR #3909, opened 2025-01-04
Added shadPS4 emulator support (by vaughngx4) #3897

vaughngx4:
- Added shadPS4 emulator definition
- Added modified version of RPCS3 import script to support PS4 games


vaughngx4:
Just adding this for reference.

Enhancement #3897

PR #3927, opened 2025-01-26
Added Lime3DS emulator profile

Jeshibu:


privateemu:
I saw this was merged and added to Playnite, however lime3DS' exe isn't lime3ds-gui.exe like Citra, it's just lime3ds.exe

PR #4038, opened 2025-06-01
Add close button to Explorer Panel

darklinkpower:

Preview:

gQHby05OgZ

Another option was to place it above but I didn't like how it left a lot of space unused for this simple button so I discarded the idea:
GdE8EMWcaw


JosefNemec:
Can you please vertically center the close button with source selection combobox?

darklinkpower:
> Can you please vertically center the close button with source selection combobox?

Sure, done. I hadn't liked that idea either before because in my opinion it looked like it would close/hide the combobox when the button is aligned next to it but on second thought I think it looks weird being aligned to the top.

Anyway, here's how it looks now:

image

PR #4040, opened 2025-06-02
Add Ymir Saturn Emulator Support

MuntKaicho:
Adds Ymir Saturn Emulator Support


darklinkpower:
Just as a heads up, you also need to add your emulator Id to its corresponding platform in `Emulation/Platforms.yaml`. An example is available in this other PR: https://github.com/JosefNemec/Playnite/pull/4050/files

Issue #3795, opened 2024-08-11
Prevent full-screen mode losing its focus when exiting a game


Starbuck1991:
### Feature description

I use a Windows HTPC frontend called Flex to launch several game launchers, such as Steam and Playnite. However, while Steam’s Big Picture mode works flawlessly (gaining focus, executing a game, exiting the game, regaining focus, launching another game, etc.), Playnite loses focus when closing a game. As a result, it remains running but hidden beneath the frontend, and it cannot be controlled until pressing Alt+Tab. Unfortunately, this behavior negates the usefulness of it's full-screen mode operated by a game controller.

JosefNemec:
We already refocus Playnite after game exits, if it doesn't work properly, then some other app is likely preventing Playnite from getting focus properly. You can read more about it [here](https://api.playnite.link/docs/manual/gettingStarted/helpAndTroubleshooting/troubleshooting.html#controller-input-doesnt-work-in-fullscreen-mode-after-exiting-a-game).

kcajjones:
I have this exact issue when running Playnite fullscreen and launching Retroarch v1.9.1 (64bit) then closing Retroarch and Playnite returns the window to fullscreen from minimized but does not regain focus. I can't see how Retroarch could be interfering with it once closed?

Starbuck1991:
> I can't see how Retroarch could be interfering with it once closed?

Exactly. Besides, no other frontend (Steam, GOG, you name it) shows this behavior under the same conditions. Ergo, something must be happening with Playnite and only with Playnite.

PR #4159, opened 2025-10-28
Fix window restoration in Fullscreen mode

DenisionSoft:
### Summary

Enforced windows restoration in Fullscreen when games exit by directly calling RestoreWindow on the Window.

### Case

In my installation, admittedly modified with many addons (though not that many, and only the most popular ones), Playnite Fullscreen window doesn't get restored after I close a game, staying minimized. This degrades the experience when playing with a gamepad, especially when streaming (i.e. via Apollo + Sunshine) when Alt+Tab or a mouse pointer are hardly available.

As far as I can tell, no issue directly mentioning this behaviour with a more traditional setup like mine exists. There is #3795, and this fix may or may not help with that issue as well.

### Proposed fix

Change `FullscreenAppViewModel.RestoreWindow()` to directly call `Window.RestoreWindow()` instead of relying on global window tracking. This should cover for more cases of window configurations, addons and other software potentially interfering.

### Demo

A build of Playnite with these changes was tested with same configuration and on the same setup, alongside the current release. The modified build shows successful Fullscreen window restoration under same conditions, allowing for seamless gamepad usage after game exit back in Playnite. Tested on Windows 11 on a latest update. Only the gamepad was used during these tests, with a mouse once to restore the window manually in the release version.

#### Before:

https://github.com/user-attachments/assets/6ecdbd1b-afd6-4b67-8e47-d716d8d23546

#### After:

https://github.com/user-attachments/assets/18d36866-148c-449d-818b-bfe31875e13b

### DIFF
--- a/source/Playnite/Native/winuser.cs
+++ b/source/Playnite/Native/winuser.cs
@@ -339,7 +339,7 @@ namespace Playnite.Native
         SHOWWINDOW = 0x0040,
         TOPMOST = NOACTIVATE | NOOWNERZORDER | NOSIZE | NOMOVE | NOREDRAW | NOSENDCHANGING
     }
-    
+
     public enum MonitorOptions : uint
     {
         MONITOR_DEFAULTTONULL = 0x00000000,


babizhu:
I think I've encountered the same issue, which is caused by the splash screen extension. 

The LastActiveWindow gets set to null, resulting in the failure of WindowManager.LastActiveWindow?.RestoreWindow();.

 My solution is to manually set WindowManager.LastActiveWindow = Application.Current.MainWindow as MainWindow;

madman-asunder:
I might have to try using playnite as my launcher again. I love it so much but I've never not had issues getting focus back.

PR #4162, opened 2025-11-01
Fix: Initialize collections in Game constructor and GetCopy

DealsBeam:
Initializes GameActions, Links, and Roms collections in the Game constructor to prevent null reference exceptions.

Also updates the GetCopy method to handle cases where these collections might be null, ensuring that the copied object always has initialized collections.

Added a new test case to verify that GetCopy correctly handles null collections.

### DIFF
--- a/source/PlayniteSDK/Models/Game.cs
+++ b/source/PlayniteSDK/Models/Game.cs
@@ -1663,6 +1663,9 @@ namespace Playnite.SDK.Models
         public Game() : base()
         {
             GameId = Guid.NewGuid().ToString();
+            GameActions = new ObservableCollection<GameAction>();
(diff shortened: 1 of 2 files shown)


JosefNemec:
Current behavior is intentional. Null and empty collections are different things with different behavior. If there are currently some places where there's null reference exception related to this, it should be handled there.

JosefNemec:
Btw was this recommended by an "AI"? Why do you have so many new PRs submitted in one hour today in multiple different projects, written in different languages and seemingly totally unrelated to each other?

JosefNemec:
Oh so you just privated your profile so people can't see that you spam repositories with AI PRs? Cool, blocked.

Issue #4068, opened 2025-07-08
Controller not working in full screen after a lengthy session


nikolaziv99:
### Bug Description

After closing the God of War: Ragnarok and going back to Playnite fullscreen, the controller stops responding altogether. It's not a focus issue as I focus Playnite manually with my mouse and keyboard, but the controller still doesn't respond. Only if I disable Controller support from the settings and reenable it does it spring back to life. If I just go into game and exit quickly, everyhing's fine. I tried reconnecting the controller but no dice. It's on, it just doesn't work in Playnite. 

### Diagnostics ID

a827a7a4-74b2-4706-b0e3-aa91e459c5a7

Jeshibu:
Try disabling the SuccessStory extension and see if it still happens

nikolaziv99:
> Try disabling the SuccessStory extension and see if it still happens

I don't have that extension installed

JosefNemec:
Looks like something messed up with SDL gamepad definitions according to the logs:

> 08-07 17:36:14.327|ERROR|FullscreenApplication:Failed to load game controller mappings:
08-07 17:36:14.327|ERROR|FullscreenApplication:Invalid RWops

Reinstall Playnite to see if it fixes the issue.

nikolaziv99:
> Looks like something messed up with SDL gamepad definitions according to the logs:
> 
> > 08-07 17:36:14.327|ERROR|FullscreenApplication:Failed to load game controller mappings:
> > 08-07 17:36:14.327|ERROR|FullscreenApplication:Invalid RWops
> 
> Reinstall Playnite to see if it fixes the issue.

Didn't help. I tried without DS4Windows, but it's still the same. I've noticed this with Sony games mostly, TLoU2 and GoW: Ragnarok

JosefNemec:
Is it 8bitdo controller?

kunit1:
This issue happens to me on a dualsense controller too. Maybe an update to SDL 3 might solve it.

mr-falafel:
Just chiming in to say I have the same experience. Boot up in fullscreen mode, play a game for 1-2 hours, exit game and gamepad no longer works in playnite. I have to use my keyboard to navigate and exit the app. I'm using an Xbox Series gamepad on Windows 11, connected via bluetooth.

mr-falafel:
[...]

PR #4174, opened 2025-11-16
Reset SDL Event loop after Playnite refocus.

kunit1:
Not sure if you want to include this fix/workaround but Playnite was unusable for me with multiple controllers. 

Fixes [4068](https://github.com/JosefNemec/Playnite/issues/4068)


JosefNemec:
I'm not sure why would this fix it, do you have some explanation/insight into why this works? The only thing this technically does is that it changes thread that calls SDL_PollEvent. I would like to know how that's related the app being minimized for long time.

kunit1:
I was too hasty in opening this. After some more testing the issue returns. I'll see if I can find a solution that doesn't involve upgrading to SDL3.

PR #4177, opened 2025-11-17
Use currently chosen display language for first start

CanRanBan:
This prevents using an unexpected language if "Windows display language" and other "Language & region" settings don't match.

### DIFF
--- a/source/Playnite/App/PlayniteApplication.cs
+++ b/source/Playnite/App/PlayniteApplication.cs
@@ -326,7 +326,7 @@ namespace Playnite
                 {
                     if (!AppSettings.FirstTimeWizardComplete)
                     {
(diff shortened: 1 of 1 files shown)


Shadas4free:
I want to pull new request but i do not know why i cannot do it. Anyway. Your new version for Updated image handling library is working wery well. After installing this update my playnite crashed down. WHen i started again it works BUT many and many i mean almost every game lost icons and also background images. So right now i must download them again. Logos are also gone. Very good work Nemec. Next time try to implement AUTOMATIC BACKUP OF WHOLE USER FILES before updating.

JosefNemec:
@Shadas4free What are you on about? How is that related to this PR in any way? Open issue if you think you found a bug.

PR #4184, opened 2025-11-23
Massive library update speed improvement

vrokolos:
Update libraries in parallel


vrokolos:
Use something like this in P11

PR #4214, opened 2025-12-27
Autofill sorting name when saving game edit window

Jeshibu:


source/Playnite.DesktopApp/ViewModels/GameEditViewModel.cs
```diff
@@ -921,6 +925,15 @@ List<Guid> consolidateIds(SelectableDbItemList selectionList, List<Guid> origina
                     game.OverrideInstallState = EditingGame.OverrideInstallState;
                 }
 
+                if (appSettings.GameSortingNameAutofill && string.IsNullOrEmpty(game.SortingName))
+                {
+                    var sortingName = sortableNameConverter.Convert(game.Name);
+                    if (sortingName != game.Name)
```
JosefNemec: Can SortableNameConverter change casing? I didn't check the code, but if so, then I would do case insensitive comparison here instead.

source/Playnite.DesktopApp/ViewModels/GameEditViewModel.cs
```diff
@@ -921,6 +925,15 @@ List<Guid> consolidateIds(SelectableDbItemList selectionList, List<Guid> origina
                     game.OverrideInstallState = EditingGame.OverrideInstallState;
                 }
 
+                if (appSettings.GameSortingNameAutofill && string.IsNullOrEmpty(game.SortingName))
```
JosefNemec: Also this should be testing against `EditingGame` not `game`, since user might have changed sorting name during this edit session. Actually it will need to test against both in case this is multi game edit.

source/Playnite.DesktopApp/ViewModels/GameEditViewModel.cs
Jeshibu: @JosefNemec I'm not checking `EditingGame` further down because it's already been set here. I guess it could also be moved into this `if` statement.

PR #4291, opened 2026-04-09
add almost all changes from Extra Emulator Profiles

Jeshibu:
This is basically all of Extra Emulator Profiles. Few changes:

- Less Switch emulators listed with the platform just to not recommend them (commented in platforms.yaml).
- Removed two TODO comments from RetroArch: `ep128emu` and `m2000` - didn't get around to figuring out how they work
- Removed an added `-nogui` from PCSX2. Don't think it's necessary when fullscreen is already a command line switch that's used. Maybe there's a use case where it makes sense though? Don't feel strongly either way.

One thing that probably deserves some explanation is the `-d` addition to shadPS4: it selects the default instance of shadPS4 to launch your game. The QT launcher can host multiple instances now. Tested again to confirm it works with 0.15.

### DIFF
--- a/source/Playnite/Emulation/Emulators/Azahar/emulator.yaml
+++ b/source/Playnite/Emulation/Emulators/Azahar/emulator.yaml
@@ -1,0 +1,9 @@
+Id: azahar
+Name: Azahar
+Website: 'https://azahar-emu.org/'
+Profiles:
+  - Name: Default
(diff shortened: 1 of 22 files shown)


Jeshibu:
Does it make sense to delist the plugin if this is merged? I guess it might still be useful for people stuck on 10.20 because of Windows 7?

PR #4305, opened 2026-04-28
Update FadeImage.xaml to display true 4k image backgrounds

RoberRaik:
Removed <UserControl.CacheMode> bracket to display full resolution background images.
Now the 4k background images in desktop mode and fullscreen mode displays in native resolution.

### DIFF
--- a/source/Playnite/Controls/FadeImage.xaml
+++ b/source/Playnite/Controls/FadeImage.xaml
@@ -6,12 +6,6 @@
              mc:Ignorable="d" 
              Name="ControlRoot"             
              d:DesignHeight="450" d:DesignWidth="800">
-    
-    <UserControl.CacheMode>
-        <BitmapCache EnableClearType="False" 
-                     RenderAtScale="1" 
-                     SnapsToDevicePixels="False" />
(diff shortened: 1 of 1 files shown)


JosefNemec:
CacheMode is there currently to improve performance because this otherwise performs badly in conjunction with blur and animation transition.

JosefNemec:
I have plans to handle this better in P11, but this is no go for P10 builds.

RoberRaik:
> CacheMode is there currently to improve performance because this otherwise performs badly in conjunction with blur and animation transition.

I understand the CacheMode is there for performance, especially when blur is enabled.

However, when blur is disabled, the BitmapCache still remains active and the control continues to render from the cached bitmap.

With high resolution backgrounds (e.g. 4K), having RenderAtScale="1" causes a noticeable loss of quality in the final static image, even when no blur is being used.

Would it make sense to conditionally enable CacheMode only when blur is active or in between transitions, and disable it otherwise? That way performance is preserved when blur is enabled or transitions are being done, while avoiding the quality degradation when it’s not needed.

PR #4314, opened 2026-05-16
Add search modes support (Normal, Fuzzy, Regex) to different components

darklinkpower:
Adds support for multiple search modes across SearchBox and filtering logic with support to switch between modes by clicking a button:

- Normal search
- Fuzzy search
- Regex search

**Changes:**
- Introduced a reusable `TextMatcher` utility to centralize matching behavior and ensure consistent search logic across different component across Playnite.
- Implemented a three-state search mode toggle button directly in the SearchBox control to allow users to quickly switch search behavior without opening settings.
- Added support to Game Edit Dialog dropdowns as well.

**TODO:**

- ~~Replace placeholder strings with proper localization strings~~ Done
- ~~Remove now obsolete global fuzzy search setting and localization entries~~ Done
- ~~Test in Fullscreen Mode~~ Done
- ~~Add/update tests (Unless they are not needed?  😁)~~ Done
- ~~Implement same functionality in Keyboard Launcher~~ Canceled
- ~~Reuse `TextMatcher` in Keyboard Launcher for consistent behavior~~ Canceled

**Edit:** There's too many moving parts in the Keyboard Launcher and I don't want to break something. I've canceled adding it, at least as part of this PR.

### DIFF
--- a/source/Playnite.DesktopApp/Controls/DdItemListSelectionBox.cs
+++ b/source/Playnite.DesktopApp/Controls/DdItemListSelectionBox.cs
@@ -171,6 +171,14 @@ namespace Playnite.DesktopApp.Controls
                     this,
                     nameof(ItemsList) + "." + nameof(ItemsList.SearchText),
                     BindingMode.TwoWay);
+
+                BindingTools.SetBinding(
+                    TextSearchBox,
+                    SearchBox.SearchModeProperty,
+                    this,
+                    nameof(ItemsList) + "." + nameof(ItemsList.SearchMode),
+                    BindingMode.TwoWay,
+                    delay: 100);
             }
 
             UpdateTextStatus();
--- a/source/Playnite.DesktopApp/Controls/SearchBox.cs
+++ b/source/Playnite.DesktopApp/Controls/SearchBox.cs
@@ -1,5 +1,6 @@
 ﻿using Playnite.Common;
 using Playnite.DesktopApp.ViewModels;
+using Playnite.SDK.Models;
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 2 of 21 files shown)


darklinkpower:
### Screenshots

Demo:

https://github.com/user-attachments/assets/5d7be6a2-6df4-4bb3-889b-2d47f9e1e1ff

Top Panel:

image

Filter Panel Name:

image

Filter Panel Version:

image

Game Edit window:

image

darklinkpower:
Tests added

image

darklinkpower:
Updated tests results after all changes

image

image

darklinkpower:
I've replaced the `^` prefix used by the Explorer Panel with a more explicit `[NameInitial]:` in https://github.com/JosefNemec/Playnite/pull/4314/commits/d050dd8e892e9733d809a440bac1879a2edabb95 to prevent collisions with valid queries that use this prefix but let me know if that's an issue. It didn't occur to me that leaving it as is could be problematic with the new Regex functionality.

But I can't find any indication of `!` being used anywhere, is it really used somewhere?

image

From what I see it's not and could be safely removed. Only use using this behavior is in this test but nothing in the real application:

image

darklinkpower:
I've changed the Explorer Panel Name prefix filter to `^:` to make it simpler. I think it keeps it simple and unlikely to cause conflicts. If that's not good and it needs to be a single character, perhaps we could use more uncommon characters such as `~`, `@` or `%` but in my opinion it works fine this way.

image

JosefNemec:
I didn't check the code yet, but quick things:
- needs to be based on `devel` branch
- I'm not a fan how it visually looks, it should be probably embedded in the search box

PR #4316, opened 2026-05-16
Group Games by first meaningful character in name

darklinkpower:
This improves title grouping for games with leading special characters. Games like `>observer_`, `[REDACTED]` and `.hack//G.U.`  currently are grouped in the `#` section that is intended for numbers. With this change they are now grouped under the first letter in their name while games starting with numbers continue to use `#`. In my opinion this is more natural.

**Before:**

Playnite DesktopApp_i2azOAW2Ge

**After:**

image

image

image

### DIFF
--- a/source/PlayniteSDK/Models/Game.cs
+++ b/source/PlayniteSDK/Models/Game.cs
@@ -2371,11 +2371,21 @@ namespace Playnite.SDK.Models
             {
                 return '#';
             }
-            else
+
+            foreach (var c in nameMatch)
             {
-                var firstChar = char.ToUpper(nameMatch[0]);
-                return char.IsLetter(firstChar) ? firstChar : '#';
+                if (char.IsDigit(c))
+                {
+                    return '#';
+                }
+
+                if (char.IsLetter(c))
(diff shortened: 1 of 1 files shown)


HerrKnarz:
Playnite already groups by sorting name and that one already removes those characters from the name of the game. Since Playnite automatically assigns sorting names for new games, this shouldn't be needed. It also would take away the option to deliberately sort using special characters if a user chooses to do so.

darklinkpower:
> Playnite already groups by sorting name and that one already removes those characters from the name of the game. Since Playnite automatically assigns sorting names for new games, this shouldn't be needed. It also would take away the option to deliberately sort using special characters if a user chooses to do so.

Automatic sorting names assignment is an optional feature, rather than an always-enabled one so the problem is that it only works if you make use of this feature. If needed, maybe a good middle ground would be to implement this new behavior only for games without a SortingName while implementing this new behavior for games without it.

HerrKnarz:
> implement this new behavior only for games without a SortingName

Yeah, that would work, IMO!

PR #4359, opened 2026-07-30
Fix locale-aware CJK font fallback in Fullscreen mode

### LINKED ISSUE #3710: Full screen mode menus display simplified chinese text incorrectly
### Bug Description

When switching to full screen mode, some of the text in the menu appears strange and does not conform to Simplified Chinese writing habits. I have marked some of the problematic text in the images.

I suspect that the font settings in this mode are incorrect and that the correct Simplified Chinese font is not being used. I suggest changing it to the system default font Microsoft YaHei, or keeping it consistent with the font in the _settings.

The reason for this problem may be the different writing habits of Chinese characters in different countries and regions. Japanese or other regional Chinese character writing may have been used incorrectly here.

### To Reproduce

_No response_

### Diagnostics ID

debbc080-9a4d-4583-a959-d5b5d7479907

### Screenshots

QQ截图20240317233257
QQ截图20240317233328
QQ截图20240317233238

RevelioLog:
Fixes #3710

## Problem

Simplified Chinese text in Fullscreen mode can be rendered with regional Han glyph variants that do not match Simplified Chinese.

Fullscreen themes use Titillium Web, which contains no CJK glyphs. WPF therefore has to select a fallback font. However, `FrameworkElement.Language` defaults to `en-US`, and Playnite did not update it when a localized UI language was loaded. WPF consequently selected and shaped fallback glyphs without the active Playnite locale.

Microsoft documents that [`FrameworkElement.Language`](https://learn.microsoft.com/en-us/dotnet/api/system.windows.frameworkelement.language) defaults to `en-US`, inherits to descendant elements, and that [`xml:lang` affects font fallback](https://learn.microsoft.com/en-us/dotnet/desktop/wpf/advanced/wpf-globalization-and-localization-overview).

## Change

- Initialize `WindowBase.Language` from `Localization.ApplicationLanguageCultureInfo`.
- Add a `zh-CN` regression test verifying that a window uses the selected application culture.

All 19 Fullscreen windows derive from `WindowBase`, so the language value is inherited by their content. This deliberately avoids hard-coding Microsoft YaHei or any other locale-specific font and also works for other languages and themes.

## Validation

- The new regression test fails before the fix (`zh-CN` expected, `en-US` actual).
- The regression test passes after the fix.
- `LocalizationTests`: 2/2 passed.
- Full `Playnite.sln` Debug/x86 build passed, including Fullscreen and test projects.
- Full `Playnite.Tests`: 279/283 passed locally. The four remaining failures are environment-dependent and unrelated to this change: one fixed checksum affected by Git CRLF conversion, and three process tests that assume English Windows output and no existing Notepad processes.

### DIFF
--- a/source/Playnite/Controls/WindowBase.cs
+++ b/source/Playnite/Controls/WindowBase.cs
@@ -15,6 +15,7 @@ using System.Windows.Automation.Peers;
 using System.Windows.Controls;
 using System.Windows.Input;
 using System.Windows.Interop;
+using System.Windows.Markup;
 using System.Windows.Media;
 
 namespace Playnite.Controls
@@ -219,6 +220,7 @@ namespace Playnite.Controls
 
         public WindowBase() : base()
         {
+            Language = XmlLanguage.GetLanguage(Localization.ApplicationLanguageCultureInfo.IetfLanguageTag);
             emptyAutomationPeer = new EmptyWindowAutomationPeer(this);
             Style defaultStyle = (Style)Application.Current?.TryFindResource(typeof(WindowBase));
             if (defaultStyle != null)
--- a/source/Playnite/Controls/WindowBase.cs
+++ b/source/Playnite/Controls/WindowBase.cs
@@ -220,7 +220,7 @@ namespace Playnite.Controls
 
         public WindowBase() : base()
         {
-            Language = XmlLanguage.GetLanguage(Localization.ApplicationLanguageCultureInfo.IetfLanguageTag);
+            Language = XmlLanguage.GetLanguage(Localization.ApplicationLanguageCultureInfo.Name);
             emptyAutomationPeer = new EmptyWindowAutomationPeer(this);
             Style defaultStyle = (Style)Application.Current?.TryFindResource(typeof(WindowBase));
             if (defaultStyle != null)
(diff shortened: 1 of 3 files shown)


JosefNemec:
Docs for `IetfLanguageTag` property say that it's deprecated and shouldn't be used.

RevelioLog:
Thanks, you're right. I replaced the deprecated `CultureInfo.IetfLanguageTag` property with `CultureInfo.Name` in c068a07.

`Name` provides the culture name expected by `XmlLanguage`. I verified that it produces the same value and round-trips through `XmlLanguage` for all 44 Playnite localization cultures, including `zh-CN` and `zh-TW`.

After the change, the full Debug/x86 solution build passes, the regression test passes, and `LocalizationTests` pass 2/2.

JosefNemec:
I will have to revert this since it causes this issue to happen: https://github.com/JosefNemec/Playnite/issues/4385

I'm not sure why, but it breaks ComboBox binding by SelectedValuePath. Affects theme selection dropdown as well.

JosefNemec:
Tried using `ConverterCulture` on that specific binding to force specific culture, didn't fix it.

PR #4362, opened 2026-08-04
Fix game list confirm button ignoring the A/B swap setting

### LINKED ISSUE #3706: Swap confirmation/cancelation binding does not work as expected
### Bug Description

In Playnite Fullscreen on WIndows 11, the "Swap confirmation/cancelation button binding" option under Input settings correctly swaps the A/B controller functions for everything *except* for when selecting games from the list of cover tiles. To select a game from the (filtered) list, I have to press B on my controller (incorrectly acting as confirm), and then on the game details screen pressing B again returns to the list (acting as cancel as expected). Occurs when using a Xbox Series controller or 8bitdo Lite.

### To Reproduce

_No response_

### Diagnostics ID

6cebd9d7-56fe-4fea-a523-d290eed0ead3

### Screenshots

_No response_

kanjieater:
The main game grid bound its select/open-details actions to a hardcoded physical A button instead of the live confirm binding, so enabling "Swap confirmation/cancellation button binding" left the list out of sync with every other screen in fullscreen mode (details, menus, dialogs), which already track the swap correctly.

Now the list follows GameControllerGesture.ConfirmationBinding and rebuilds its bindings when the setting changes, and the on-screen button hints update to match.

Fixes #3706

I know things are generally on pause because majority of code base is being rewritten for Playnite 11 but was hoping a small safe pr like this could taken still as the button mapping has been causing me havoc lately 😆

### DIFF
--- a/source/Playnite.FullscreenApp/Controls/Views/Main.cs
+++ b/source/Playnite.FullscreenApp/Controls/Views/Main.cs
@@ -101,12 +101,6 @@ namespace Playnite.FullscreenApp.Controls.Views
 
             this.mainModel.AppSettings.Fullscreen.PropertyChanged += Fullscreen_PropertyChanged;
             this.mainModel.PropertyChanged += MainModel_PropertyChanged;
-            GameControllerGesture.ConfirmationBindingChanged += GameControllerGesture_ConfirmationBindingChanged;
-        }
-
-        private void GameControllerGesture_ConfirmationBindingChanged(object sender, EventArgs e)
-        {
-            SetListCommandBindings();
         }
 
         private void MainModel_PropertyChanged(object sender, PropertyChangedEventArgs e)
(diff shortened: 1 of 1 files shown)


source/Playnite.FullscreenApp/Controls/Views/Main.cs
```diff
@@ -101,6 +101,12 @@ public Main(FullscreenAppViewModel mainModel) : base()
 
             this.mainModel.AppSettings.Fullscreen.PropertyChanged += Fullscreen_PropertyChanged;
             this.mainModel.PropertyChanged += MainModel_PropertyChanged;
+            GameControllerGesture.ConfirmationBindingChanged += GameControllerGesture_ConfirmationBindingChanged;
```
JosefNemec: This should be done via checking settings value instead, like it's done with `SwapStartDetailsAction` on line 148. You'll just check for `SwapConfirmCancelButtons` also.

kanjieater: Done, thanks

PR #4383, opened 2026-08-30
Fixes for description focus and spacing

siegmour:
### Description

Several small fixes to how game descriptions render and behave in Fullscreen mode in both the stock and custom themes. They address separate problems but overlap in the same layout, so they're intended to be taken together.

### Changes

**HtmlTextView.cs**
- `moreButton` used `Hidden` rather than `Collapsed`, reserving layout space on every description short enough not to need the button.
- `moreButton` had a hardcoded 5px bottom margin. Being inside the scroll viewer's extent, it shifted the text when scrolling to the bottom. It had no style of its own, so themes could only reach it via an implicit `Button` style affecting every plain Button. It now looks up "LoadMoreButton", falling back to the implicit style if a theme doesn't define it.
- Block-level elements kept their default bottom margin. The renderer does not collapse the last one, so descriptions ending in a paragraph or list reported more height than they rendered, leaving empty scrollable space.

**DescriptionView.html** — the same CSS rule. Needed in both, since a theme shipping its own template overrides the built-in one.

**ScrollViewerEx.cs** — `ScrollChanged` moved focus on any scroll reaching the bottom, including one caused by `MakeVisible` revealing a newly focused child. The Load More button was therefore unreachable whenever another
focusable element followed the description. Now suppressed only for scrolls that follow focus moving into the content, so scrolling to the bottom still advances focus as before.

**Button.xaml** — adds a `LoadMoreButton` style with the spacing, applied only to that button. Spacing previously came from the description's trailing block margin, so it varied by formatting and was absent for plain text.

The `moreButton` margin removal and the `ScrollViewerEx` change depend on each other; the margin and the default theme not having anything selectable below `moreButton` were masking the focus bug.

Tested in Fullscreen with descriptions that fit, that scroll, that end in lists or plain text, and that exceed 10,000 characters. Tested in both default theme, and custom theme with different layouts.

### Screenshots

Here are some example screenshots which highlight some of the fixes in the default theme, although they don't cover all the fixes above:

### Before fixes:

Screenshot 2026-08-30 095729
Screenshot 2026-08-30 223903
Screenshot 2026-08-30 224056

### After fixes:

Screenshot 2026-08-30 224313
Screenshot 2026-08-30 223943
Screenshot 2026-08-30 224047

### DIFF
--- a/source/Playnite.DesktopApp/Themes/Desktop/Default/DefaultControls/Button.xaml
+++ b/source/Playnite.DesktopApp/Themes/Desktop/Default/DefaultControls/Button.xaml
@@ -42,4 +42,8 @@
             </Trigger>
         </Style.Triggers>
     </Style>                  
+
+    <Style x:Key="LoadMoreButton" BasedOn="{StaticResource {x:Type Button}}" TargetType="Button">
+        <Setter Property="Margin" Value="0,5,0,5" />
+    </Style>               
 </ResourceDictionary>
--- a/source/Playnite.FullscreenApp/Controls/ScrollViewerEx.cs
+++ b/source/Playnite.FullscreenApp/Controls/ScrollViewerEx.cs
@@ -12,6 +12,7 @@ namespace Playnite.FullscreenApp.Controls
 {
     public class ScrollViewerEx : ScrollViewer
     {
+        private bool focusMovedWithin;
         public double CustomScrollAmount { get; set; } = 0;
 
         static ScrollViewerEx()
@@ -29,12 +30,19 @@ namespace Playnite.FullscreenApp.Controls
         {
             ScrollChanged += ScrollViewerEx_ScrollChanged;
             PreviewKeyDown += ScrollViewerEx_PreviewKeyDown;
+            GotKeyboardFocus += ScrollViewerEx_GotKeyboardFocus;
         }
 
         private void ScrollViewerEx_Unloaded(object sender, RoutedEventArgs e)
         {
             ScrollChanged -= ScrollViewerEx_ScrollChanged;
             PreviewKeyDown -= ScrollViewerEx_PreviewKeyDown;
+            GotKeyboardFocus -= ScrollViewerEx_GotKeyboardFocus;
+        }
+
+        private void ScrollViewerEx_GotKeyboardFocus(object sender, KeyboardFocusChangedEventArgs e)
+        {
+            focusMovedWithin = e.NewFocus is DependencyObject newFocus && IsAncestorOf(newFocus);
         }
 
         private void ScrollViewerEx_PreviewKeyDown(object sender, KeyEventArgs e)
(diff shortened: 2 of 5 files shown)


siegmour:
### Changes addressing the issues

**ScrollViewerEx.cs**

>That ScrollViewerEx.cs change will break navigation in other places where scroll viewer is used.

I have updated the implementation to be more targeted so it only affects the bug. Existing behaviour is unchanged everywhere else — scrolling to the bottom still advances focus on the same press. The suppressed case is a scroll caused by focus moving into the content, which is the bug.

Noting again that the `VerticalOffset == 0` branch appears to have the same issue but I cannot reproduce it with the current content, so I've left it alone. It can be addressed for consistency by changing `if (VerticalOffset == 0)` to `if (!focusJustMoved && VerticalOffset == 0)`

Please correct me if I'm wrong - my previously suggested implementation doesn't break the navigation. It indeed does change the behavior slightly, the focus won't move automatically but will instead require a key press once you're at the last scrollable line. Personally I prefer the behavior without an auto advance, but of course this is subjective so I leave the decision to you.

**Button.xaml**

>And that Button.xaml change affects all buttons, not just a button in description views, that's also no go.

The proper approach and my preference would be to assign a Style to the Load More button and target that specifically. This also has the benefit that theme designers now have a hook for it. This is exactly what I've done with the new commits.

For context on the previous version, I had checked and moreButton is the only plain button in the current codebase without an explicit style, so the setter on the base style didn't reach anything else in practice. I was prioritizing a minimal diff, but a keyed style is the better shape regardless.

**DescriptionView.html**

> That HTML template change breaks descriptions

This seems to affect only descriptions where plain text follows a block element directly, without <br> or a wrapping tag. Since only margin-bottom is reset, spacing between two block elements still comes from the following element's margin-top — it's only missing when followed by plain text. None of the descriptions I tested were formatted like that, and that's why I missed it.

I think `:last-child` would scope the rule to the trailing element and solve this cleanly, but it doesn't match in the renderer. Looking at the current HTML-Renderer source it does appear to be implemented (LastChildSelector, registered in PseudoClassSelectorFactory). I can't tell if references/HtmlRenderer.dll in Playnite predates it or is a modified build without it.

Is there another way to scope this that I'm missing?

This change does address a real-world issue with descriptions. however I assume updating the HTML-Renderer would fall out of the scope for Playnite 10, with Playnite 11 coming up. This is the least important change in this PR, as it's something that can be easily implemented custom theme side, so I can drop it if you prefer.

JosefNemec:
I'm not going to accept any changes that result in breaking behavior. Which the template change and scrollviewer are. Especially the scrollviwer one can be problematic to 3r party themes that could be relying on current behavior.

JosefNemec:
> This seems to affect only descriptions where plain text follows a block element directly, without
or a wrapping tag. 

My example on Discord was from real description from GOG. So this would at least break rendering of GOG descriptions. I also tested the behavior in Firefox for the same HTML, and it renders the same way as current Playnite. So it should imo stay like it is now.

JosefNemec:
> I think :last-child would scope the rule to the trailing element and solve this cleanly, but it doesn't match in the renderer.

The control supports CSS 2 only. `last-child` is CSS 3 feature. We are using custom fork from here: https://codeberg.org/CrowIsTaken/HTML-Renderer

The original control was archived for long time, but I now checked and they seem to be developing it again. You probably checked the original repo that might have had the feature added.

It's probably out of scope for P10 to look into rebasing on that new work, but I'll look into it for P11.

